PR #3, opened 2018-12-11
Fixed negative times in format_bustime

chrusher:


common/common.py
```diff
@@ -50,10 +50,11 @@ def format_bustime(bustime, round="millisecond"):
 		110:50
 		159:59:59.999
 		-10:30:01.100
-	Note that a negative value only indicates the number of hours after the start
-	is negative, the number of minutes/seconds is simply time past the hour.
-	eg. the bustime "-01:20:00" indicates the run begins in 40 minutes, not 80 minutes.
```
ekimekim: Probably still good to have the docstring clarify the behaviour here, eg:
```
Negative times are formatted as time-until-start, preceeded by a minus sign.
eg. "-1:20:00" indicates the run begins in 80 minutes.
```

PR #5, opened 2018-12-25
Initial work on restreamer

ekimekim:
Provides some APIs for serving segments, and generating a playlist.

### DIFF
--- a/common/common.py
+++ b/common/common.py
@@ -2,7 +2,14 @@
 """A place for common utilities between wubloader components"""
 
 
+import base64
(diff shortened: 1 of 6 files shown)


ekimekim:
I've been trying to stick to `datetime.datetime` and `datetime.timedelta` objects for time data everywhere.

chrusher:
Ah ok, I still suggest being more explicit in the docstring what types the
inputs for functions should be.

C

ekimekim:
nod. will do

PR #12, opened 2018-12-29
Chrusher/backfiller

chrusher:
Prototype of the backfiller


chrusher:
> We might want to support a limit on how long ago can be backfilled, eg. so we can have 'partial copies' of the stream on backup nodes, or to download a specific time range.

Yeah this will be easy to do. The `backfill.backfill` function right now supports not only backfilling all hours but also only backfilling the last N hours or a specific list of hours. You could easily do a specific time range using `common.hour_paths_for_range`.

`backfill.main` right now is very much a rough prototype. I fully expect this part of the backfiller to change dramatically as we develop the `wubloader`.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,249 @@
...
+	# based on restreamer.list_segments
+	# could just call restreamer.list_segments but this avoids http/json
+	# overheads
+	path = os.path.join(base_dir, stream, variant, hour)
+	try:
+		local_segments = [name for name in os.listdir(path) if not
+						name.startswith('.')]
+	except OSError as e:
+		if e.errno != errno.ENOENT:
+			raise:
```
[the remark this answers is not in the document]
chrusher: Doh! Fixed.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,249 @@
...
+		else:
```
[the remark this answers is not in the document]
chrusher: Dropped the else

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,249 @@
...
+
+def list_local_segments(base_dir, stream, variant, hour):
+
+	# based on restreamer.list_segments
+	# could just call restreamer.list_segments but this avoids http/json
+	# overheads
+	path = os.path.join(base_dir, stream, variant, hour)
+	try:
+		local_segments = [name for name in os.listdir(path) if not
+						name.startswith('.')]
```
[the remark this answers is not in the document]
chrusher: Good to know. Normally I'm bad at breaking lines even when they are way too long.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,249 @@
...
+
+	hours = resp.json()
+
+	return hours
+
+def list_remote_segments(node, stream, variant, hour, timeout=TIMEOUT):
+
+	# just a wrapper around a call to restreamer.list_segments
+
+	uri = 'https://{}/files/{}/{}/{}'.format(node, stream, variant, hour_str)
```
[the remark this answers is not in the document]
chrusher: I've changed node to be the full part.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,249 @@
...
+
+	with open(temp_name, 'w') as f:
+		for chunk in resp.iter_content(8192):
+			f.write(chunk)
+
+	dir_path = os.path.join(base_dir, stream, variant, hour)
+
+	if not os.path.exists(dir_path):
+		try:
+			os.mkdir(dir_path)
```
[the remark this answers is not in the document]
chrusher: I've moved ensure_directory to common.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,249 @@
...
+	try:
+		resp = requests.get(uri, stream=True, timeout=timeout)
+
+	except requests.exceptions.Timeout:
+		return False
+
+	if resp.status_code != requests.codes.ok:
+		return False
+
+	temp_name = 'temp_backfill'
```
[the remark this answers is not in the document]
chrusher: Done 

I've also moved rename from the downloader to common and used it rather than just the os.rename.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,249 @@
...
+	
+	# just a wrapper around a call to restreamer.list_hours
+
+	uri = 'https://{}/files/{}/{}'.format(node, stream, variant)
+
+	try:
+		resp = requests.get(uri, timeout=timeout)
+
+	except requests.exceptions.Timeout:
+		return []
```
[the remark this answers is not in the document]
chrusher: I've taken out all the exception handling and put it in backfill

chrusher:
I've done most of what you've asked.

I've moved the exception handling to `backfill` which now just calls `get_nodes` then loops over the nodes, giving up on a node if an exception is raised.

I've not put much effort into improving `backfill` or `main` right now since they will likely be completely rewritten to backfill in parallel and to be more flexible. On this note I didn't try an implement a shorter callback for nodes that fail. I did add the jitter in the sleep by moving `downloader.jitter` to `common` though.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+import os
+import time
+import datetime
+import errno
+import uuid
+
+import requests
+
+import common
+
```
ekimekim: I try to leave double-newlines between major "blocks" such as each top-level function or class, and between imports, constants, and the first function/class. So with that in mind: extra newline here, and between each function below

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+HOUR_FMT = '%Y-%m-%dT%H'
+TIMEOUT = 5
```
ekimekim: should add a comment saying what the timeout is for eg. "timeout on all requests to remote servers"

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+
+def get_nodes():
+
+	# either read a config file or query the database to get the addresses
+	# of the other nodes
+	# figure out some way that the local machine isn't in the list of returned
+	# nodes so that 
+
+	# as a prototype can just hardcode some addresses.
+	# each element in nodes is a 'protocol://host:port/' string
```
ekimekim: turn this comment block into docstring

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+	return nodes
+
+def list_local_segments(base_dir, stream, variant, hour):
+
+	# based on restreamer.list_segments
+	# could just call restreamer.list_segments but this avoids http/json
+	# overheads
+	path = os.path.join(base_dir, stream, variant, hour)
+	try:
+		local_segments = [name for name in os.listdir(path) if not name.startswith('.')]
```
ekimekim: just directly return the value instead of assigning it to a variable to be returned later

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+	# of the other nodes
+	# figure out some way that the local machine isn't in the list of returned
+	# nodes so that 
+
+	# as a prototype can just hardcode some addresses.
+	# each element in nodes is a 'protocol://host:port/' string
+
+	nodes = []
+
+	return nodes
```
ekimekim: obviously this function will be changed later, but if i were critiquing it for style in its current form I'd say just `return []` without an intermediate variable.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+
+def list_local_segments(base_dir, stream, variant, hour):
+
+	# based on restreamer.list_segments
+	# could just call restreamer.list_segments but this avoids http/json
+	# overheads
```
ekimekim: last word here can just go on prev line

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+
+	# as a prototype can just hardcode some addresses.
+	# each element in nodes is a 'protocol://host:port/' string
+
+	nodes = []
+
+	return nodes
+
+def list_local_segments(base_dir, stream, variant, hour):
+
```
ekimekim: such short functions generally don't need a line of whitespace at the start like this

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+		local_segments = []
+
+	return local_segments
+
+def list_remote_hours(node, stream, variant, timeout=TIMEOUT):
+	
+	# just a wrapper around a call to restreamer.list_hours
+	uri = '{}/files/{}/{}'.format(node, stream, variant)
+	resp = requests.get(uri, timeout=timeout)
+	hours = resp.json()
```
ekimekim: avoid intermediate variable: `return resp.json()`

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+	return hours
+
+def list_remote_segments(node, stream, variant, hour, timeout=TIMEOUT):
+
+	# just a wrapper around a call to restreamer.list_segments
+	uri = '{}/files/{}/{}/{}'.format(node, stream, variant, hour_str)
```
ekimekim: hour_str is not defined here. you probably just want hour

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+	except OSError as e:
+		if e.errno != errno.ENOENT:
+			raise
+		
+		local_segments = []
+
+	return local_segments
+
+def list_remote_hours(node, stream, variant, timeout=TIMEOUT):
+
```
ekimekim: as above, these small functions can do without the whitespace between the definition and first line

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+	return remote_segments
+
+# based on _get_segment in downloader/main
+# very basic error handling
+def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
+			timeout=TIMEOUT):
+
+
+	path = os.path.join(base_dir, stream, variant, hour, missing_segment)
+	if os.path.exists(path):
```
ekimekim: should have a comment here explaining why we double-check if it exists

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+		return
+
+	substrs = path.split('-')
```
ekimekim: This will break when the hash contains a `-`, which it often does (we're using base64 with A-Z, a-z, 0-9 and _ and -).
It will also break if stream or variant contain a -, which i'm not sure if that's possible.

I've been considering abstracting out all path-related ops into a single spot in common to ensure they stay unified. I'm still not sure if it's worth the extra indirection. In the meantime a more correct treatment would be:
```
dir_name = os.path.dirname(path)
date, duration, _ = os.path.basename(path).split('-', 2)
temp_name = "-".join([date, duration, "temp", str(uuid.uuid4())])
temp_path = os.path.join(dir_name, "{}.ts".format(temp_name))
```

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+	temp_path = '-'.join(substrs[:-1] + [str(uuid.uuid4()) + '.st'])
+	common.ensure_directory(temp_path)
+
+	uri = '{}/segments/{}/{}/{}/{}'.format(node, stream, variant, hour, missing_segment)
+	resp = requests.get(uri, stream=True, timeout=timeout)
```
ekimekim: should check response status with `resp.raise_for_status()` before trying to access iter_content

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+
+	with open(temp_path, 'w') as f:
```
ekimekim: should wrap all this (from temp_path creation to rename) in a:
```
try:
  ...
except Exception:
  if os.path.exists(temp_path):
    os.remove(temp_path)
  raise
```
to ensure the temp file is deleted on error

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+		for chunk in resp.iter_content(8192):
+			f.write(chunk)
+
+	common.rename(temp_path, path)
+
```
ekimekim: too many newlines between functions

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+def backfill(base_dir, stream, variants, hours=None, nodes=None):
+	
+	# loop over nodes backfilling from each
+
+	if nodes is None:
+		nodes = get_nodes()
+
+	#ideally do this in parallel
+	for node in nodes:
+
```
ekimekim: too many newlines here. try to only put newlines mid-function to split up logical "blocks" of doing-stuff. in a function as short as this I would probably not put any.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+def backfill_node(base_dir, node, stream, variants, hours=None, recent_cutoff=60):
+
+
+	# if hours is int, backfill last hours hourdirs
+	# else if hours is None, backfill all hourdirs
+	# else assume hours is iterable and backfill those hourdirs
+
+
+	if isinstance(hours, int):
```
ekimekim: `isinstance(int)` is a tricky thing...for one thing, you at least should include both `int` and `long`. Then there's problems like people accidentially passing in `1.0` when they meant `1` and getting unexpected behaviour (or Fraction(1), or Decimal(1)). There's just a lot of "number-like" things in python and they tend to (in most circumstances) expect to be treated as int if they're integral.

Generally I try to avoid the issue by using more easily-distingishable types and falling back on int as the base case.
So in this scenario i would check None first, then check if it's iterable:
```
def is_iterable(x):
  try:
    iter(x)
  except TypeError:
    return False
  return True
```
then finally fall back to number.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+
+		now = datetime.datetime.utcnow()
+		hours = [(now - i * timedelta(hours=1)).strftime(HOUR_FMT) for i in range(n_hours)]
+
+	for variant in variants:
+
+		if hours is None:
+			node_hours = list_remote_hours(node, stream, variant)
+		else:
+			node_hours = hours
```
ekimekim: rather than creating a new variable `node_hours` that is either `hours` or a default value, I would just modify hours:
```
if hours is None:
  hours = ...
```

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+
+		for hour in node_hours:
+
+			local_segments = list_local_segments(base_dir, stream, variant, hour)
+			local_segments = set(local_segments)
```
ekimekim: i'd probably just include the `set(...)` on the line above

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+			remote_segments = list_remote_segments(node, stream, variant, hour)
+			remote_segments = set(remote_segments)
+			missing_segments = remote_segments - local_segments
+
+			for missing_segment in missing_segments:
+
+				#to avoid getting in the downloader's way ignore segments less than recent_cutoff old
+				time_str = '{}:{}'.format(hour, missing_segment.split('-')[0])
+				segment_time = datetime.datetime.strptime(time_str, HOUR_FMT + ':%M:%S.%f')
```
ekimekim: huh. for some reason i had it in my head that datetime.strptime couldn't use `%f`, but i just tried it and it works. i should probably go back and eliminate unnessecary usage of `dateutil.parser.parse` (it's still needed for the HLS timestamps, which may have timezone information that needs to be corrected for before producing a timezone-less UTC datetime)

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+			time.sleep(common.jitter(60 * sleep_time))
+
+
+	
+
+
+
+	
+
+
```
ekimekim: so many newlines!

common/common.py
```diff
@@ -252,3 +252,36 @@ def best_segments_by_start(hour):
...
+	by hash anyway, so if the filepath already exists the file itself is already there.
+	In this case, we delete the source file.
+	"""
+	try:
+		os.rename(old, new)
+	except OSError as e:
+		if e.errno != errno.EEXIST:
+			raise
+		os.remove(old)
+
```
ekimekim: extra newline here

common/common.py
```diff
@@ -252,3 +252,36 @@ def best_segments_by_start(hour):
...
+	if os.path.exists(dir_path):
+		return
+	ensure_directory(dir_path)
+	try:
+		os.mkdir(dir_path)
+	except OSError as e:
+		# Ignore if EEXISTS. This is needed to avoid a race if two getters run at once.
+		if e.errno != errno.EEXIST:
+			raise
+
```
ekimekim: and here

downloader/downloader/main.py
```diff
@@ -470,7 +451,7 @@ def _get_segment(self):
 					logging.warning("Got 403 Forbidden for segment, giving up: {}".format(self.segment))
 					return
 				resp.raise_for_status()
-				self.ensure_directory(temp_path)
+				ensure_directory(temp_path)
```
ekimekim: missing `common.`

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,200 @@
...
+# when starting the backfiller and every few hours, backfill everything
+# more frequently, backfill the last couple hours
+# (last three hour directories so always at least two hours).
+
+import os
+import time
+import datetime
+import errno
+import uuid
```
ekimekim: each block of imports should be sorted alphabetically.

ekimekim:
oh, i just remembered: you should ensure you check that you're not grabbing any `-temp-UUID.ts` files that might exist. they'd be harmless, but a waste of space/bandwidth. not to mention they generally exist for so short a time they'd just 404.

we could alternately filter those out at the server end, but i feel like that kind of logic makes more sense to have here in the spirit of loose coupling. right now the restreamer just serves all files and leaves it up for whatever clients to decide which they want.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
+			timeout=TIMEOUT):
+	"""Get a segment from a node.
+
+	Fetches stream/variant/hour/missing_segment from node and puts it in base_dir/stream/variant/hour/missing_segment. If the segment already exists locally, this does not attempt to fetch it."""
+
+	path = os.path.join(base_dir, stream, variant, hour, missing_segment)
+	# check to see if file already exists to avoid unnecessarily copying it
```
ekimekim: expand comment a little: we already checked once when we got a list of "missing segments", why are we checking again? eg. `file might have been created since we listed our segments, check again to see if we don't need to copy it after all`

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,227 @@
...
+	"""Backfill from remote node.
+
+	Backfill from node/stream/variants to base_dir/stream/variants.
+
+	Keyword arguments:
+	hours -- If None (default), backfill all available hours. If iterable, backfill only hours in iterable. Otherwise backfill the last N hours, starting with the lastest.
+	start -- Only backfill hours starting after or equal to this datetime object. If None (default), backfill all hours.
+	stop -- Only backfill hours starting before or equal to this datetime object. If None (default), backfill all hours.
+	order -- If 'random', randomise the order of hours. If 'forward', sort the hours in acceding order. If 'reverse', sort the hours in descending order. Otherwise, do not change the order of hours (default).
+	recent_cutoff -- Skip backfilling segments younger than this number of seconds to prioritise letting the downloader grab these segments."""
```
ekimekim: wrap some of the longer lines here. wrapped lines should be indented, eg:
```
hours: If blah blah blah something long
    and also some more text
start: next line blah blah
```

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,227 @@
...
+
+	if hours is None:
+		hours = list_remote_hours(node, stream, variant)
+	elif is_iterable(hours):
+		None
```
ekimekim: use `pass` for an empty block, perhaps with a comment, eg.
```
    pass # hours is already in desired format
```

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,227 @@
...
+		stop=None, order=None, recent_cutoff=60):
+	"""Backfill from remote node.
+
+	Backfill from node/stream/variants to base_dir/stream/variants.
+
+	Keyword arguments:
+	hours -- If None (default), backfill all available hours. If iterable, backfill only hours in iterable. Otherwise backfill the last N hours, starting with the lastest.
+	start -- Only backfill hours starting after or equal to this datetime object. If None (default), backfill all hours.
+	stop -- Only backfill hours starting before or equal to this datetime object. If None (default), backfill all hours.
+	order -- If 'random', randomise the order of hours. If 'forward', sort the hours in acceding order. If 'reverse', sort the hours in descending order. Otherwise, do not change the order of hours (default).
```
ekimekim: typo: ascending

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,227 @@
...
+
+	if nodes is None:
+		nodes = get_nodes()
+
+	#ideally do this in parallel
+	for node in nodes:
+		try:
+			backfill_node(base_dir, node, stream, variants, hours, start, stop, order=order)
+		except Exception as e:
+			print node, e
```
ekimekim: use `logging.exception()` to log the error, eg. `logging.exception("Error while backfilling node {}".format(node))`. logging.exception will automatically include the traceback and error info.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,227 @@
...
+	if start is not None:
+		hours = [hour for hour in hours if hour >= start]
+	if stop is not None:
+		hours = [hour for hour in hours if hour <= stop]
+
+	# useful if running in parallel and expecting to fetch a segments from
+	# multiple hours (say on start up) so that you don't try to backfill the
+	# same hour at the same time 
+	if order == 'random':
+		hours = random.shuffle(hours)
```
ekimekim: bug here. `random.shuffle` returns None, and shuffles the list in-place. stupid, i know.

backfiller/backfiller/main.py
```diff
@@ -0,0 +1,227 @@
...
+		for hour in hours:
+
+			local_segments = set(list_local_segments(base_dir, stream, variant, hour))
+			remote_segments = set(list_remote_segments(node, stream, variant, hour))
+			missing_segments = remote_segments - local_segments
+
+			for missing_segment in missing_segments:
+
+				#ignore temporary files
+				if 'temp' in missing_segment:
```
ekimekim: this will cause any segment whose hash happens to contain the letters 'temp' to not be replicated.
You should probably use `common.parse_segment_path()` instead, and simply ignore any segments that don't parse correctly (gives a ValueError). This can also replace the `.ts` check and the segment time parsing below.

ekimekim: Just for fun and because I thought this was amusing, I ran the numbers on the chance of this occuring:

Let's consider how many hashes contain "temp" as a substring.
Due to how base64 works, and that we're representing 256 bits in 43 characters,
the last character has a bit form of xxxx00, ie one of: AEIMQUYcgkosw048,
which does not include p. So last 4 chars can't be "temp", so start char
of the temp substring must be in the first 38 chars.
So there are 38 positions for "temp". For each one, the other 38 unrestricted chars + 1 final char can be anything, which is 64^38 * 16.
So there are 38 * 16 * 64^38 = (19 * 2) * 2^4 * (2^6)^38 = 19 * 2^233 hashes containing
the substring "temp". There are of course 2^256 hashes total, so the chance of a hash
containing "temp" is 19 * 2^233 / 2^256 = 19 / 2^23 or 0.00023%.
The chance of N hashes NOT containing "temp" in any of them is (1 - 19/2^23)^N.
If we expect a week's worth of segments at 6s per segment, that's 100800 segments,
giving us a chance of 79.59% of hitting this in practice.

ekimekim: Actually it's a little less likely because i didn't consider things that contain "temp" twice, so eg. a hash like "temptemp000000..." would be double-counted.

ekimekim:
btw, when you go to merge this, please flatten out all the merges from master into a linear history. you can do this automatically by clicking the merge drop down and selecting "rebase and merge", or on the command line by running `git rebase master`.

PR #13, opened 2018-12-30
restreamer: Also have an endpoint for generating cut videos on demand

ekimekim:
This is mainly just for testing until we get the database and proper cutter up,
but it might prove useful to have in the long run too.

This code will probably end up being totally rewritten,
as it uses the most naive form of cutting and reencoding,
and it has a whole bunch of http-serving specifics intertwined with the cutting logic.

### DIFF
--- a/restreamer/restreamer/main.py
+++ b/restreamer/restreamer/main.py
@@ -5,11 +5,13 @@ import functools
 import json
 import logging
 import os
+import shutil
 import signal
 
 import dateutil.parser
 import gevent
-from flask import Flask, url_for, request, abort
+from flask import Flask, url_for, request, abort, Response
+from gevent import subprocess
 from gevent.pywsgi import WSGIServer
 
 from common import get_best_segments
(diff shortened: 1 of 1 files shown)


restreamer/restreamer/main.py
```diff
@@ -163,6 +168,96 @@ def generate_media_playlist(stream, variant):
...
+	"""
+	start = dateutil.parser.parse(request.args['start'])
+	end = dateutil.parser.parse(request.args['end'])
+	if end <= start:
+		return "End must be after start", 400
+
+	allow_holes = request.args.get('allow_holes', 'false').lower()
+	if allow_holes not in ["true", "false"]:
+		return "allow_holes must be one of: true, false", 400
+	allow_holes = allow_holes == "true"
```
chrusher: I had to reread this a couple times to make sure it wasn't a coding error. I'd toss a pair of parentheses around the comparison to make it more clear that you are assigning the result of the comparison.

PR #20, opened 2019-01-02
Chrusher/backfiller 

chrusher:
Added a Dockerfile for the backfiller and added the backfiller to the build file.

I was able to build and run the backfiller container locally. The backfiller behaved as expected for an empty list of nodes.

Also, added more logging to the backfiller as well as adding the ability to order or shuffle both hours and the missing segments in `backfiller.backfill_node`.


chrusher:
With respect with having too many options for `backfiller_node` I've removed everything except the segment order since we can do everything else by what we pass to hours and restricting the backfill within an hour doesn't seem all that useful.

backfiller/backfiller/main.py
```diff
@@ -230,19 +209,26 @@ def backfill_node(base_dir, node, stream, variants, hours=None, start=None,
...
 
 							
-def main(base_dir='/mnt', stream='desertbus', variants=['source'], fill_wait=5, full_fill_wait=180, sleep_time=1):
+def main(base_dir='.', stream='', variants='', fill_wait=5, full_fill_wait=180, sleep_time=1):
 	"""Prototype backfiller service.
 
-	Do a backfill of the last 3 hours from stream/variants from all nodes initially before doing a full backfill from all nodes. Then every sleep_time minutes check to see if more than fill_wait minutes have passed since the last backfill. If so do a backfill of the last 3 hours. Also check whether it has been more than full_fill_wait minutes since the last full backfill; if so, do a full backfill."""
+	Do a backfill of the last 3 hours from stream/variants from all nodes
+	initially before doing a full backfill from all nodes. Then every sleep_time
+	minutes check to see if more than fill_wait minutes have passed since the		last backfill. If so do a backfill of the last 3 hours. Also check whether		it has been more than full_fill_wait minutes since the last full backfill;
```
ekimekim: this line still needs a newline in it

PR #31, opened 2019-01-08
Performance improvements for get_best_segments

ekimekim:
Approx 80% of our time for large get_best_segments calls were in date parsing.
This uses strptime instead, which is much faster.

We also ensure we let other things run while we're processing, to prevent stalling
all other work / causing the service to go unresponsive.

This PR is based on #30 and will need to be rebased once that's merged

### DIFF
--- a/common/common/segments.py
+++ b/common/common/segments.py
@@ -11,8 +11,6 @@ import os
 import sys
 from collections import namedtuple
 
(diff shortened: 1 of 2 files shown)


chrusher:
Ekimekim: Can you clean up the merge/rebase please?

PR #36, opened 2019-01-22
Add a doc covering how the database is used

ekimekim:
I fully expect the exact list of sheet inputs, edit inputs and outputs to change.
The important thing I wanted to codify here was the state machine and the behaviour of the cutters.


ekimekim:
missing cols:
challenge submitter / auction winner text sheet input
poster moment boolean sheet input

ekimekim:
add state `DONE` for when yt has finished transcoding to all qualities, rename `UPLOADED` to `TRANSCODING`
allow `FINALIZING` to go to `TRANSCODING` _or_ `DONE` since not all upload locations will have the concept

PR #37, opened 2019-02-26
Chrusher/backfiller/events

chrusher:
Moved the backfiller over to be gevent based to backfill from multiple nodes at once.

### DIFF
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -36,29 +36,6 @@ def encode_strings(o):
 		return o.encode('utf-8')
 	return o
 
-def get_nodes(node_source=None):
-	"""List address of other wubloaders.
-	
-	This returns a list of the other wubloaders as strings of the form
-	'protocol://host:port/'"""
-
-
-	if node_source == 'file':
-		logging.info('Fetching list of other nodes from file')
-		# read a config file
-		# should I allow any filename?
-		pass
-
-	elif node_source == 'database':
-		logging.info('Fetching list of other nodes from database')
-		# query the database
-		pass
-
-	else:
-		nodes = []
-	
-	return nodes
-
 
 def list_local_segments(base_dir, stream, variant, hour):
 	"""List segments in a given hour directory.
@@ -171,7 +148,7 @@ class BackfillerManager(object):
 
 	NODE_INTERVAL = 300 #seconds between updating list of nodes
 
-	def __init__(self, base_dir, stream, variants, static_nodes=[], start=None, run_once=False, node_source=None):
+	def __init__(self, base_dir, stream, variants, static_nodes=[], start=None, run_once=False, node_file=None, node_database=None):
 		"""Constructor for BackfillerManager.
 		Creates a manager for a given stream with specified variants. If
 		static_nodes is None, manager"""
@@ -181,7 +158,8 @@ class BackfillerManager(object):
 		self.static_nodes = static_nodes
 		self.start = start
 		self.run_once = run_once
-		self.node_source = node_source
+		self.node_file = node_file
+		self.node_database = node_database
 		self.stopping = gevent.event.Event()
 		self.logger = logging.getLogger("BackfillerManager({})".format(stream))
 		self.workers = {} # {node url: worker}
@@ -207,7 +185,7 @@ class BackfillerManager(object):
 	def run(self):
 		self.logger.info('Starting')
 		while not self.stopping.is_set():
-			new_nodes = set(get_nodes(self.node_source) + self.static_nodes)
+			new_nodes = set(self.get_nodes())
 			exisiting_nodes = set(self.workers.keys())
 			to_start = new_nodes - exisiting_nodes
 			for node in to_start:
@@ -227,6 +205,30 @@ class BackfillerManager(object):
 		for worker in self.workers.values():
 			worker.done.wait()
 
+	def get_nodes(self):
+		"""List address of other wubloaders.
+		
+		This returns a list of the other wubloaders as strings of the form
+		'protocol://host:port/'"""
+		nodes = self.static_nodes + []
+	
+		if self.node_file is not None:
+			self.logger.info('Fetching list of nodes from {}'.format(self.node_file))
+			try:
+				with open(self.node_file) as f:
+					for line in f.readlines():
+						substrs = line.split()
+						if not len(line) or substrs[0][0] == '#':
+							continue
+						nodes.append(substrs[0])
+			except IOError:
+				self.logger.info('{} not found'.format(self.node_file))
+	
+		if self.node_database:
+			self.logger.info('Fetching list of nodes from {}'.format(self.node_database))
+			# query the database
+	
+		return nodes
 
 class BackfillerWorker(object):
 	"""Backfills segments from a node.
@@ -328,12 +330,13 @@ class BackfillerWorker(object):
 		
 		self.logger.info('Worker stopped')
 		self.done.set()
-		del self.manager.workers[self.node]
+		if self.node in self.manager.workers:
+			del self.manager.workers[self.node]
 
 							
 def main(base_dir='.', streams='', variants='', metrics_port=8002,
 	static_nodes='', backdoor_port=0, start=None, run_once=False,
-	node_source=None):
+	node_file=None, node_database=None):
 	"""Backfiller service."""
 
 	streams = streams.split(',') if variants else []
@@ -356,7 +359,7 @@ def main(base_dir='.', streams='', variants='', metrics_port=8002,
 	workers = []
 	for stream in streams:
 		logging.info('Starting backfilling {} with {} as variants to {}'.format(stream, ', '.join(variants), base_dir))
-		manager = BackfillerManager(base_dir, stream, variants, static_nodes, start, run_once, node_source)
+		manager = BackfillerManager(base_dir, stream, variants, static_nodes, start, run_once, node_file, node_database)
 		managers.append(manager)
 		workers.append(gevent.spawn(manager.run))
 
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -1,5 +1,4 @@
 """Download segments from other nodes to catch stuff this node missed."""
-# TODO more logging
 
 import datetime
 import errno
@@ -7,8 +6,10 @@ import logging
 import os
 import random
 import signal
+import socket
 import uuid
 
+import argh
 import dateutil.parser
 import gevent.backdoor
 import prometheus_client as prom
@@ -26,15 +27,6 @@ segments_backfilled = prom.Counter(
 
 HOUR_FMT = '%Y-%m-%dT%H'
 TIMEOUT = 5 #default timeout in seconds for remote requests 
-
-def encode_strings(o):
-	if isinstance(o, list):
-		return [encode_strings(x) for x in o]
-	if isinstance(o, dict):
-		return {k.encode('utf-8'): encode_strings(v) for k, v in o.items()}
-	if isinstance(o, unicode):
-		return o.encode('utf-8')
-	return o
 
 
 def list_local_segments(base_dir, stream, variant, hour):
@@ -175,6 +167,9 @@ class BackfillerManager(object):
 
 	def start_worker(self, node):
 		"""Start a new worker for given node."""
+		if self.stopping.is_set():
+			logging.debug("Refusing to create new worker because we're stopping")
+			return
 		worker = BackfillerWorker(self, node)
 		assert node not in self.workers, "Tried to start worker for node {!r} that already has one".format(node)
 		self.workers[node] = worker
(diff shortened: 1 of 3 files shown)


backfiller/backfiller/main.py
```diff
@@ -129,159 +131,263 @@ def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
...
+	
+			for hour in hours:
+	
+				self.logger.info('Backfilling {}/{}'.format(variant, hour))
+	
+				local_segments = set(list_local_segments(self.base_dir, self.stream, variant, hour))
+				remote_segments = set(list_remote_segments(self.node, self.stream, variant, hour))
+				missing_segments = list(remote_segments - local_segments)
+	
+				# useful if running in parallel so multiple nodes don't request the same segment at the same time
```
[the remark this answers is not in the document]
chrusher: The idea with the random order is to lessen the odds that two workers request the same segment at the same time. I think it is more likely that one node will be requesting an hour from a multiple nodes at the same time (for example on start up) than multiple nodes requesting the same hour from a single node at the same. If multiple workers are trying to backfill the same hour at the same time and the segments are requested in order, the workers will request the same segment at the same time, leading to duplicated effort.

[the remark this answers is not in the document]
ekimekim: Hm. I prefer requesting things in order because it makes it clearer to the operator what the backfiller is doing / where it's up to. But not downloading the same thing twice is also a nice property. I guess just update the comment to make it clearer?

chrusher:
Yeah, having the two lengths of backfills should be overkill most of the time. I can think of a couple places where it could be an issue. At the end of a long run (DBfH usually lasts ~160 hours) if you are backfilling the entire run, it is a decent number of hours to backfill (over 300 assuming two versions of the stream). Depending on how long it takes for a single hour to be done this could start to be a long amount of time. Based on my testing with two nodes on the same LAN, a backfill without fetching any segments takes about 100 ms which would be fine (a full backfill would be ~30 s) but if each hour takes significantly longer the total time would also start to become unmanageable. Real life testing of the system across widely separated nodes will provide us with how long it can take to backfill. 

The other issue comes if you spin up a completely new node at the end of the week. The first backfill will take a long time to copy all the hours. However, since the backfiller ignores the most recent 60 s to prevent competing with the downloader, some of the most recent segments will get missed the first time around and will potentially not get backfilled for a while. Again, real life testing will reveal how long it takes to backfill a full hour worth of segments.

An easy fix to both issues to limit the backfills to the last N hours while retaining the ability to manually backfill older segments as needed.

chrusher:
I've tried to address your comments and suggestions in the pushed changes.

I've simplified the behaviour backfiller considerably to only have one backfiller mode. I've added some options to `main()` to make up for this though. First, the backfiller can be run only once via `run_once`. Second, the option `start` allows you to limit the backfills to either the last N hours if `start` is a number or to hours since a time if `start` is a datetime.

I've added error handling code to the workers. If a backfill fails, the worker will retry while geometrically backing off.

I've rewritten `get_nodes()` to allow the list of nodes to set when `main()` is called (via the option `static_nodes`), fetched from a file (`node_file`) or from a database (`node_database`). These are not exclusive and the first two work (the database one is just a stub for now). I've already found reading the nodes from a file to be quite useful for testing.

The backfiller now handles multiple channels via multiple managers (via code copied from the downloader).

backfiller/backfiller/main.py
```diff
@@ -24,7 +25,7 @@
 
 
 HOUR_FMT = '%Y-%m-%dT%H'
-TIMEOUT = 5 #default timeout for remote requests 
+TIMEOUT = 5 #default timeout in seconds for remote requests 
 
 def encode_strings(o):
```
ekimekim: nitpick: this is also defined in `common` and we should just use that one

ekimekim: in fact, this definition is already unused and can just be deleted

backfiller/backfiller/main.py
```diff
@@ -129,159 +114,308 @@ def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
...
+		while not self.stopping.is_set():
+
+			try:
+				self.backfill(list_hours(self.node, self.stream, self.variants, self.start))
+				self.failures = 0 #reset failure count on a successful backfill
+				if not self.run_once:
+					self.stopping.wait(common.jitter(self.WAIT_INTERVAL))
+
+			except Exception:
+				self.failures += 1
```
ekimekim: `self.failures` is only ever used locally to this function, it probably makes more sense as a local variable.

backfiller/backfiller/main.py
```diff
@@ -129,159 +114,308 @@ def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
...
+				delay = common.jitter(self.RETRY_INTERVAL * 2**self.failures)
```
ekimekim: this expontential backoff needs an upper limit, or it will take forever to retry after a few rounds. I suggest something like one minute (reached after 4 failures).

backfiller/backfiller/main.py
```diff
@@ -129,159 +114,308 @@ def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
...
 
-		now = datetime.datetime.now()
+							
+def main(base_dir='.', streams='', variants='', metrics_port=8002,
+	static_nodes='', backdoor_port=0, start=None, run_once=False,
+	node_file=None, node_database=None, localhost=None):
+	"""Backfiller service.
 
-		if now - full_fill_start > datetime.timedelta(minutes=full_fill_wait):
+	Keyword arguments:
```
ekimekim: I know this is slightly more annoying to write, but if you take all this help text and write it like this instead:
```
@argh.arg("--base-dir", help="Directory to which segments will...")
@argh.arg("--streams", help="Streams to backfill. Comma seperated if multiple.")
```
then they'll show up in the USAGE if you run `backfiller --help` etc, which is very helpful.

backfiller/backfiller/main.py
```diff
@@ -129,159 +114,308 @@ def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
...
+		self.done.set()
+		if self.node in self.manager.workers:
+			del self.manager.workers[self.node]
 
-		now = datetime.datetime.now()
+							
+def main(base_dir='.', streams='', variants='', metrics_port=8002,
+	static_nodes='', backdoor_port=0, start=None, run_once=False,
+	node_file=None, node_database=None, localhost=None):
+	"""Backfiller service.
```
ekimekim: `variants` should default to `source` like downloader.

assuming we go with having localhost refer to the node's "name", it would make sense to default to `socket.gethostname()`.

`streams` should probably be a positional arg, not an option `--streams`, since it's almost always required (backfilling no streams is possible but not useful).
Ideally, it should work the same way as downloader and can be specified as multiple positional args, ie. `backfiller --blah=blah channel1 channel2 channel3`. You can do this by making `streams` the first arg, not giving it a default, and adding:
```
@argh.arg("streams", nargs="*")
```
the `nargs="*"` tells it that the positional arg may be given 0 or more times (also possible: `?` for optional, `N` for exactly n times (eg. `1`, which is the default), or `+` for "1 or more").

backfiller/backfiller/main.py
```diff
@@ -129,159 +114,308 @@ def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
...
+		self.localhost = localhost
+		self.stopping = gevent.event.Event()
+		self.logger = logging.getLogger("BackfillerManager({})".format(stream))
+		self.workers = {} # {node url: worker}
+
+	def stop(self):
+		"""Shut down all workers and stop backfilling."""
+		self.logger.info('Stopping')
+		for node in self.workers.keys():
+			self.stop_worker(node)
```
ekimekim: I think we should leave it up to the `run()` function to stop the workers.
My concern is that if `stop()` is called while `run()` is halfway through, say, fetching nodes, then `run()` is now going to START those workers again (since they're no longer in `self.workers`), plus other similar issues. The best way to avoid race conditions on shared state like that is to only ever mutate that state from a canonical place, which in this case would be `run()`. So what this would look like would be that at the end of our `run` loop, we'd stop all workers after exiting the loop.

backfiller/backfiller/main.py
chrusher: I don't see in `common` where there is a timeout.

`TIMEOUT` is used as the default value of `timeout` in `list_remote_hours`, `list_remote_segments` and `get_remote_segment`. I suppose I could just replace `timeout` in these with `TIMEOUT`.

backfiller/backfiller/main.py
ekimekim: Ok, after looking at the interaction between this, the `else: self.stop()` in `run()`, and the run_once behaviour, I think I understand.

The virtue of stopping the workers in `stop()` is that it will stop them even if we exited the `run()` loop due to `run_once`, and are currently waiting for them to finish.

The extra `stop()` in `run()` that is only called if we're already stopping is to shut down any extra errant workers caused by the race.

The race between this and the `run()` loop is probably best resolved by adding a check to `start_worker`, something like:
```
if self.stopping.is_set():
  logging.debug("Refusing to create new worker because we're stopping")
  return
```

Then we no longer need the extra `stop()` in run.


Some comments explaining this are probably also in order.

backfiller/backfiller/main.py
```diff
@@ -129,159 +114,308 @@ def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
...
+		if self.node_file is not None:
+			self.logger.info('Fetching list of nodes from {}'.format(self.node_file))
+			try:
+				with open(self.node_file) as f:
+					for line in f.readlines():
+						substrs = line.split()
+						if not len(line) or substrs[0][0] == '#':
+							continue
+						nodes.append(substrs[0])
+			except IOError:
```
ekimekim: Rather than handling errors in file or database lookups as ignoring their contents (which will cause the respective workers to be shut down), I think errors in get_nodes should simply be propagated upwards and handled in `run()` with a retry. That way if eg. the file is unavailable briefly, or the database is down, we maintain our existing set of workers instead of stopping everything.

chrusher: Ok, I've copied the same back off and number or tries from the workers to handle errors in get_nodes. If max retries is reached, existing workers are allow to keep running but no further attempts are made to call get_nodes.

backfiller/backfiller/main.py
chrusher: Fixed this by adding the check to start and added some comments.

backfiller/backfiller/main.py
chrusher: Added a limit of 4 tries

backfiller/backfiller/main.py
chrusher: Done

backfiller/backfiller/main.py
chrusher: Done

backfiller/backfiller/main.py
```diff
@@ -129,159 +114,308 @@ def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
...
+				self.logger.info('{} not found'.format(self.node_file))
+	
+		if self.node_database:
+			self.logger.info('Fetching list of nodes from {}'.format(self.node_database))
+			# query the database
+	
+		if self.localhost is not None:
+			nodes = [node for node in nodes if self.localhost not in node]
```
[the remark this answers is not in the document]
chrusher: Changed this to your suggestion.

backfiller/backfiller/main.py
```diff
@@ -129,159 +114,308 @@ def get_remote_segment(base_dir, node, stream, variant, hour, missing_segment,
...
+	manager starts one. If a worker corresponds to a node not in either list,
+	the manager stops it. If run_once, only backfill once."""
+
+	NODE_INTERVAL = 300 #seconds between updating list of nodes
+
+	def __init__(self, base_dir, stream, variants, static_nodes=[], start=None,
+			run_once=False, node_file=None, node_database=None, localhost=None):
+		"""Constructor for BackfillerManager.
+		Creates a manager for a given stream with specified variants. If
+		static_nodes is None, manager"""
```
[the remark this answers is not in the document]
chrusher: Not sure what happened here - I might have only pushed some of the comments. Current version has all of them decorated as argh.arg.

backfiller/backfiller/main.py
```diff
@@ -315,18 +325,22 @@ def backfill(self, hours, segment_order='random', recent_cutoff=60):
...
-				self.failures = 0 #reset failure count on a successful backfill
+				failures = 0 #reset failure count on a successful backfill
 				if not self.run_once:
 					self.stopping.wait(common.jitter(self.WAIT_INTERVAL))
 
 			except Exception:
-				self.failures += 1
-				delay = common.jitter(self.RETRY_INTERVAL * 2**self.failures)
+				failures += 1
+				if failures > MAX_RETRIES:
```
ekimekim: I don't think giving up after too many retries is a good idea, here or in get nodes. we want the service to be able to recover itself and it can't do that if it stops trying.
when i said we should have a limit on the backoff i meant a cap on the retry interval after which we continue to retry every minute or whatever, not give up.

PR #41, opened 2019-05-27
Update DATABASE.md

MasterGunner:
Changed allow_holes and uploader_whitelist to be edit inputs - there's no need for them to come from the sheet; and we'll have an admin dashboard for modifying them if needed.


chrusher:
upload_location should also be an edit input.

PR #44, opened 2019-06-10
cutter: Initial implementation

ekimekim:
This series of commits introduces the cutter component.

The cutter has two jobs:
* To cut videos, taking them through states EDITED -> TRANSCODING
* To monitor TRANSCODING videos for when they're complete

This initial implementation only implements uploading to youtube,
and always to the same account. Later, we may want to implement alternate "upload locations"
which could be alternate youtube accounts, or other video services.

It could also use a lot more instrumentation.

NOTE: I have attempted to break this very large PR up into a series of commits that
introduces one change at a time. It's highly recommended you review by looking at the
commits one by one.

### DIFF
--- a/cutter/cutter/youtube.py
+++ b/cutter/cutter/youtube.py
@@ -1,0 +1,111 @@
+
+import logging
+import time
+
+import gevent
+import requests
+
+
+class Youtube(object):
+	"""Manages access to youtube and maintains an active access token"""
+
+	ACCESS_TOKEN_ERROR_RETRY_INTERVAL = 10
+	# Refresh token 10min before it expires (it normally lasts an hour)
+	ACCESS_TOKEN_REFRESH_TIME_BEFORE_EXPIRY = 600
+
+	def __init__(self, client_id, client_secret, refresh_token):
+		self.logger = logging.getLogger(type(self).__name__)
+		self.client_id = client_id
+		self.client_secret = client_secret
+		self.refresh_token = refresh_token
+
+		self._first_get_access_token = gevent.spawn(self.get_access_token)
+
+	@property
+	def access_token(self):
+		"""Blocks if access token unavailable yet"""
+		self._first_get_access_token.join()
+		return self._access_token
+
+	def get_access_token(self):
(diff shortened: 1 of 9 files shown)


cutter/cutter/youtube.py
```diff
@@ -0,0 +1,107 @@
...
+
+		self._first_get_access_token = gevent.spawn(self.get_access_token)
+
+	@property
+	def access_token(self):
+		"""Blocks if access token unavailable yet"""
+		self._first_get_access_token.join()
+		return self._access_token
+
+	def get_access_token(self):
```
chrusher: Doc strings please

cutter/cutter/youtube.py
```diff
@@ -0,0 +1,107 @@
...
+				})
+				resp.raise_for_status()
+				data = resp.json()
+				self._access_token = data['access_token']
+				expires_in = (start_time + data['expires_in']) - time.time()
+				if expires_in < self.ACCESS_TOKEN_REFRESH_TIME_BEFORE_EXPIRY:
+					self.logger.warning("Access token expires in {}s, less than normal leeway time of {}s".format(
+						expires_in, self.ACCESS_TOKEN_REFRESH_TIME_BEFORE_EXPIRY,
+					))
+				gevent.spawn_later(expires_in, self.get_access_token)
```
chrusher: Does a stop in the parent Cutter or TranscodeChecker propagate throw to gevent.spawn_later()?

I might be missing something but the this seems to be called exactly at the expiry time and not the ACCESS_TOKEN_REFRESH_TIME_BEFORE_EXPIRY before the expiry.

ekimekim: It does not. However, the extra work is harmless if it does happen, and in any case it almost certainly won't since you're generally only stopping when you're bringing down the entire process. The pending greenlet will not prevent process shutdown.

You're correct, good catch. Fixing.

ekimekim:
Addressed comments.
Currently there's no fancy backoffs in here, I don't think they're worth the complexity. They are at least random jittered, but tbh I don't think I'd want these retries any faster OR any slower - faster would be spammy and slower would be slow to recover.

PR #45, opened 2019-06-12
Chrusher/thrimshim/initial

chrusher:
Initial implementation of thrimshim, the code to manage communication between the wubloader and the [thrimbletrimmer](https://github.com/MasterGunner/Thrimbletrimmer-III). It provides a simple web service to request rows from the database and to update the editing columns of the a row in the database.

### DIFF
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -7,6 +7,7 @@ import gevent
 import gevent.backdoor
 from gevent.pywsgi import WSGIServer
 import flask
+from psycopg2 import sql
 
 import common
 from common import database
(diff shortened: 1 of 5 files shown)


thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+			SELECT id, state
+			FROM events
+			WHERE id = %s;""", ident)
+	old_row = results.fetchone()
+	if old_row is None:
+		return 'Row {} not found'.format(ident), 404
+	assert old_row.id == ident
+	
+	if old_row.state not in ['UNEDITED', 'EDITED', 'CLAIMED']:
+		return 'Video already published', 400
```
ekimekim: Allow me to indulge my HTTP-nerd side here, and say that this could more specifically be a `403 Forbidden` (since we're requesting a state change on an object whose state can no longer be changed by us).

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+				'video_start', 'video_end', 'video_title', 'video_description',
+				'video_channel', 'video_quality']
+	state_columns = ['state', 'uploader', 'error', 'video_link'] 
+	columns = edit_columns + state_columns
+
+	#check edit columns are in new_row
+	row_keys = new_row.keys()
+	for column in edit_columns + ['state']:
+		if column not in row_keys:
+			return 'Missing field {} in JSON'.format(column), 400
```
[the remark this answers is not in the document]
ekimekim: Although, does it make more sense to allow partial updates? ie. if a particular column is omitted, that should be fine and we're simply not changing it.
This is especially relevant for columns whose values we'll be overriding anyway, like `state` when `video_link` is set.

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+		return 'Video already published', 400
+	
+	# handle state columns
+	# handle non-empty video_link as manual uploads
+	# otherwise clear other state columns
+	if 'video_link' in new_row and new_row['video_link']:
+		new_row['state'] = 'DONE'
+		new_row['upload_location'] = 'manual'
+	else:
+		new_row['video_link'] = None
```
ekimekim: if we're not intending to change it, we should not put `video_link` in `new_row` at all. Better to just not refer to it at all than to explicitly set it to NULL.

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+		new_row['upload_location'] = None
```
ekimekim: `upload_location` should be set by the editor here, so they can choose between the main youtube or whatever else. It should be an edit column.

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+		if new_row['state'] not in ['EDITED']:
```
ekimekim: instead of silently correcting, any value besides `EDITED` or `UNEDITED` should be an error

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+			new_row['state'] = 'UNEDITED'
+	new_row['uploader'] = None
+	new_row['error'] = None
+
+	# actually update database
+	build_query = sql.SQL("""
+		UPDATE events
+		set {}
```
ekimekim: nitpick: all caps `SET` please. just convention for SQL keywords.

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+		WHERE id = %(id)s""").format(sql.SQL(", ").join(
+			sql.SQL("{} = {}").format(
+				sql.Identifier(column), sql.Placeholder(column),
+			) for column in columns
```
ekimekim: rather than needing to ensure all columns are present in `new_row`, we should just change this line to `for column in new_row` so that everything in new_row gets updated and everything that isn't is just left alone.

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+		))
+	kwargs = {column:new_row[column] for column in columns}
+	kwargs['id'] = ident
+	with database.transaction(conn):
+		result = database.query(conn, build_query, **kwargs)
```
ekimekim: instead of building kwargs, we can just use new_row directly here: `database.query(conn, build_query, id=ident, **new_row)`

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+		if new_row['state'] not in ['EDITED']:
+			new_row['state'] = 'UNEDITED'
+	new_row['uploader'] = None
+	new_row['error'] = None
+
+	# actually update database
+	build_query = sql.SQL("""
+		UPDATE events
+		set {}
+		WHERE id = %(id)s""").format(sql.SQL(", ").join(
```
ekimekim: There's a race condition here. The row might have been modified between our previous SELECT and this update. We should also check `AND state IN ('UNEDITED', 'EDITED', 'CLAIMED')`.
Since rows are never deleted, if we then find that our UPDATE did not update any rows, we know it's because the video changed state, and can respond appropriately.

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,160 @@
...
+			sql.SQL("{} = {}").format(
+				sql.Identifier(column), sql.Placeholder(column),
+			) for column in columns
+		))
+	kwargs = {column:new_row[column] for column in columns}
+	kwargs['id'] = ident
+	with database.transaction(conn):
+		result = database.query(conn, build_query, **kwargs)
+	if result.rowcount != 1:
```
ekimekim: see above about checking state in the update query. then here we can respond with a similar error to earlier about video already being uploaded.

thrimshim/thrimshim/main.py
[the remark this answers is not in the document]
chrusher: I think ignoring fields that are irrelevant or will be overridden is the better option. The way I imagined thrimbletrimmer working is taking the row object returned by the GET method, updating what fields it need to, then sending the updated row object back to the POST method.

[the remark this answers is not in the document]
chrusher: Now, only the columns in non_null_columns and state are required

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,186 @@
...
+		AND state IN ('UNEDITED', 'EDITED', 'CLAIMED')"""
+		).format(sql.SQL(", ").join(
+			sql.SQL("{} = {}").format(
+				sql.Identifier(column), sql.Placeholder(column),
+			) for column in new_row.keys()
+		))
+	with database.transaction(conn):
+		result = database.query(conn, build_query, id=ident, **new_row)
+	if result.rowcount != 1:
+		if result.rowcount == 0.:
```
ekimekim: i assume `0.` is a typo, unless you intended to compare to float `0.0` here.

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,186 @@
...
+	if missing:
+		return 'Fields missing in JSON: {}'.format(', '.join(missing)), 400
+	#get rid of irrelevant columns
+	extras = set(new_row) - set(edit_columns + state_columns)
+	for extra in extras:
+		del new_row[extra]
+
+	conn = app.db_manager.get_conn()
+	#check a row with id = ident is in the database
+	with database.transaction(conn):
```
ekimekim: no need to wrap this in a transaction, as it's only one query

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,186 @@
...
+		results = database.query(conn, """
+			SELECT id, state 
+			FROM events
+			WHERE id = %s;""", ident)
```
ekimekim: nit: no semicolon on the query

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,186 @@
...
+		UPDATE events
+		SET {}
+		WHERE id = %(id)s
+		AND state IN ('UNEDITED', 'EDITED', 'CLAIMED')"""
+		).format(sql.SQL(", ").join(
+			sql.SQL("{} = {}").format(
+				sql.Identifier(column), sql.Placeholder(column),
+			) for column in new_row.keys()
+		))
+	with database.transaction(conn):
```
ekimekim: as above, no transaction needed for single query

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,186 @@
...
+		result = database.query(conn, build_query, id=ident, **new_row)
+	if result.rowcount != 1:
+		if result.rowcount == 0.:
+			with database.transaction(conn):
```
ekimekim: no transaction needed

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,186 @@
...
+				check_result = database.query(conn, """
+					SELECT id, state
+					FROM events
+					WHERE id = %s;""", ident)
```
ekimekim: no semicolon

thrimshim/thrimshim/main.py
```diff
@@ -0,0 +1,186 @@
...
+			sql.SQL("{} = {}").format(
+				sql.Identifier(column), sql.Placeholder(column),
+			) for column in new_row.keys()
+		))
+	with database.transaction(conn):
+		result = database.query(conn, build_query, id=ident, **new_row)
+	if result.rowcount != 1:
+		if result.rowcount == 0.:
+			with database.transaction(conn):
+				check_result = database.query(conn, """
```
ekimekim: I think this second query's completely unnessecary, and we can assume that if our update fails, it's because the video is published.

PR #46, opened 2019-06-14
database: Add constraints on edit inputs that they must be non-NULL if state != UNEDITED

ekimekim:
This should help prevent changing state to EDITED with any of these fields unset,
which would blow up the cutter.

We also fix up upload_location, which was set up as a sheet input (NOT NULL DEFAULT ''),
and add a similar constraint saying any DONE columns must have non-NULL video link.

TODO test. There may be syntax errors.


ekimekim:
Tests are failing because we never created a quay repo for `wubloader-thrimshim`. also need one for cutter. fixing.

PR #47, opened 2019-06-15
Remove all usage of dateutil except when absolutely required

ekimekim:
dateutil attempts some fuzzy matching of datetimes, uses the system's local timezone
by default (if timestamp doesn't have a timezone) and returns offset-aware datetime objects.

We don't want any of these things in most circumstances - we would rather error out on a bad
timestamp than make a guess as to its meaning, and we always want to deal strictly in UTC.

The only exception to this is when parsing incoming HLS playlists.


ekimekim:
Closing in favor of #48

PR #49, opened 2019-06-15
Fixes the state transitions of the trimbleshim to allow video links t…

chrusher:
…o be changed or removed

### DIFF
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -9,11 +9,13 @@ import gevent
 import gevent.backdoor
 from gevent.pywsgi import WSGIServer
 import prometheus_client
+import psycopg2
 from psycopg2 import sql
 
 import common
 from common import database
 
+psycopg2.extras.register_uuid()
 app = flask.Flask('thrimshim')
 
 def cors(app):
@@ -42,7 +44,6 @@ def metrics():
 @app.route('/thrimshim/<uuid:ident>', methods=['GET', 'POST'])
 def thrimshim(ident):
 	"""Comunicate between Thrimbletrimmer and the Wubloader database."""
-	ident = str(ident)
 	if flask.request.method == 'POST':
 		row = flask.request.json
 		return update_row(ident, row)
@@ -63,6 +64,7 @@ def get_row(ident):
 	assert row.id == ident
 	response = row._asdict()
 
+	response['id'] = str(response['id'])	
 	response = {
 		key: (
 			value.isoformat() if isinstance(value, datetime.datetime)
@@ -141,21 +143,30 @@ def update_row(ident, new_row):
 
 @app.route('/thrimshim/manual-link/<uuid:ident>', methods=['POST'])
 def manual_link(ident):
-	"""Manually set a video_link"""
+	"""Manually set a video_link if the state is 'UNEDITED' or 'DONE' and the 
+	upload_location is 'manual'."""
 	link = flask.request.json
 	conn = app.db_manager.get_conn()
 	results = database.query(conn, """
+		SELECT id, state, upload_location 
+		FROM events
+		WHERE id = %s""", ident)
+	old_row = results.fetchone()
+	if old_row is None:
+		return 'Row {} not found'.format(ident), 404
+	if old_row.state != 'UNEDITED' and not (old_row.state == 'DONE' and old_row.upload_location == 'manual'):
+		return 'Invalid state {} for manual video link'.format(old_row.state), 403		
+
+	results = database.query(conn, """
 		UPDATE events 
-		SET state='DONE', upload_location = 'Manual', video_link = %s
-		WHERE id = %s""", link, str(ident))
-	if results.rowcount != 1:
-		return 'Row id = {} not found'.format(ident), 404
+		SET state='DONE', upload_location = 'manual', video_link = %s
+		WHERE id = %s AND (state = 'UNEDITED' OR (state = 'DONE' AND
+			upload_location = 'manual'))""", link, ident)
 	logging.info("Row {} video_link set to {}".format(ident, link))
 	return ''	
 	
 
-
-@app.route('/thrimshim/reset/<uuid:ident>')
+@app.route('/thrimshim/reset/<uuid:ident>', methods=['POST'])
 def reset_row(ident):
 	"""Clear state and video_link columns and reset state to 'UNEDITED'."""
 	conn = app.db_manager.get_conn()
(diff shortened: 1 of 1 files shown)


thrimshim/thrimshim/main.py
```diff
@@ -79,13 +76,13 @@ def get_row(ident):
...
-
 def update_row(ident, new_row):
 	"""Updates row of database with id = ident with the edit columns in
 	new_row.
 
-	If a 'video_link' is provided in update, interperet this as a manual video
-	upload and set state to 'DONE'"""
+	If a 'video_link' is provided in update, interpret this as a manual video
+	upload and set state to 'DONE'. If state currently is 'DONE', and a empty
+	'video_link' is present, reset state to 'UNEDITED'."""
```
ekimekim: no longer true

thrimshim/thrimshim/main.py
```diff
@@ -114,11 +111,14 @@ def update_row(ident, new_row):
...
 	
-	if old_row.state not in ['UNEDITED', 'EDITED', 'CLAIMED']:
+
+	if old_row.state not in ['UNEDITED', 'EDITED', 'CLAIMED'] and not ('video_link' in new_row and new_row['video_link']):
 		return 'Video already published', 403
 	
 	# handle state columns
-	# handle non-empty video_link as manual uploads
+	# interpret non-empty video_link as manual uploads
+	# interpret state == 'DONE' and an empty video link as instructions to reset
```
ekimekim: no longer true

thrimshim/thrimshim/main.py
```diff
@@ -114,11 +111,14 @@ def update_row(ident, new_row):
 		return 'Row {} not found'.format(ident), 404
 	assert old_row.id == ident
 	
-	if old_row.state not in ['UNEDITED', 'EDITED', 'CLAIMED']:
+
+	if old_row.state not in ['UNEDITED', 'EDITED', 'CLAIMED'] and not ('video_link' in new_row and new_row['video_link']):
```
ekimekim: video link check can be written more concisely as `new_row.get('video_link')`, which will return None (and thus be false) if the key isn't present

thrimshim/thrimshim/main.py
```diff
@@ -138,15 +138,15 @@ def update_row(ident, new_row):
 	new_row['error'] = None
 
 	# actually update database
-	build_query = sql.SQL("""
+	query_str = """
```
ekimekim: Use `sql.SQL().format()`, not `string.format()`, for formatting SQL queries. String formatting makes it way too easy to accidentially put in an SQL injection.

thrimshim/thrimshim/main.py
```diff
@@ -138,15 +138,15 @@ def update_row(ident, new_row):
...
 		UPDATE events
-		SET {}
+		SET {{}}
 		WHERE id = %(id)s
-		AND state IN ('UNEDITED', 'EDITED', 'CLAIMED')"""
-		).format(sql.SQL(", ").join(
-			sql.SQL("{} = {}").format(
-				sql.Identifier(column), sql.Placeholder(column),
-			) for column in new_row.keys()
+		{}""".format("AND state IN ('UNEDITED', 'EDITED', 'CLAIMED')" if not ('video_link' in new_row and new_row['video_link']) else "")
```
ekimekim: rather than modifying the query directly, you can just do something like `AND state IN %(allowed_states)s`

thrimshim/thrimshim/main.py
```diff
@@ -155,6 +155,25 @@ def update_row(ident, new_row):
 	logging.info('Row {} updated to state {}'.format(ident, new_row['state']))
 	return ''
 
+@app.route('/thrimshim/reset/<ident>')
+def reset_row(ident):
+	"""Clear state and video_link columns and reset state to 'UNEDITED'."""
+	try:
```
ekimekim: you can get flask to do this bit for you by making your route have `<ident:uuid>`.

thrimshim/thrimshim/main.py
```diff
@@ -155,6 +155,25 @@ def update_row(ident, new_row):
...
+		uuid.UUID(ident)
+	except ValueError:
+		return 'Invalid format for id', 400
+	conn = app.db_manager.get_conn()
+	results = database.query(conn, """
+		UPDATE events 
+		SET STATE='UNEDITED', error = NULL, video_id = NULL, video_link = NULL,
```
ekimekim: style nit: lowercase for state

thrimshim/thrimshim/main.py
```diff
@@ -155,6 +155,25 @@ def update_row(ident, new_row):
...
+		uploader = NULL
```
ekimekim: style nit: indent second line of the SET clause to make it clear it's still part of the SET.

thrimshim/thrimshim/main.py
chrusher: Annoyingly, I can't seem to pass uuid objects to psycopg2 so I need to turn them into strings

ekimekim: Looks like you need to register it on startup: http://initd.org/psycopg/docs/extras.html?highlight=uuid#psycopg2.extras.register_uuid

I'm sure there's some annoying backwards-compatability reason for the default to be broken.

thrimshim/thrimshim/main.py
```diff
@@ -147,14 +130,46 @@ def update_row(ident, new_row):
...
 
+
+@app.route('/thrimshim/manual-link/<uuid:ident>', methods=['POST'])
+def manual_link(ident):
+	"""Manually set a video_link"""
+	link = flask.request.json
+	conn = app.db_manager.get_conn()
+	results = database.query(conn, """
+		UPDATE events 
+		SET state='DONE', upload_location = 'Manual', video_link = %s
```
ekimekim: let's make `manual` lowercase, just because having one consistent rule is easier to remember

thrimshim/thrimshim/main.py
```diff
@@ -147,14 +130,46 @@ def update_row(ident, new_row):
...
+		WHERE id = %s""", link, str(ident))
```
ekimekim: this should only work when `state = 'UNEDITED' OR (state = 'DONE' AND upload_location = 'Manual')`, ie. either this is the first edit, or the previous edit was already a manual upload.
reset flow should cover everything else

thrimshim/thrimshim/main.py
```diff
@@ -147,14 +130,46 @@ def update_row(ident, new_row):
...
+	if results.rowcount != 1:
+		return 'Row id = {} not found'.format(ident), 404
+	logging.info("Row {} video_link set to {}".format(ident, link))
+	return ''	
+	
+
+
+@app.route('/thrimshim/reset/<uuid:ident>')
```
ekimekim: method=POST

thrimshim/thrimshim/main.py
chrusher: Ok, so in the cases when we need to replace a youtube upload that has been blocked we will need call both reset and manual-link

PR #53, opened 2019-06-18
Proposal for security model

MasterGunner:
Mostly a dump of thoughts on how to approach authentication and user permissions, vaguely based on past discussions. Will need some work and fleshing out.

### DIFF
--- a/SecurityModel.md
+++ b/SecurityModel.md
@@ -13,20 +13,25 @@ only have a single update event every few minutes, we can authenticate only as n
 
 ### Wubloader Authorization
 
-Authorized tasks via Thrimshim will be:
-* `update_row`: Updates row in the Events table. Compares incoming updates against the `event_column_whitelist` for the role associated with the current user, to limit which fields can be updated by a user/role.
-* `manual_link`: Override the `video_link` field in the Events Table, in case of a manual upload.
-* `reset_row`: Clear `state` and `video_link` columns and reset `state` to 'UNEDITED' in Events table.
+Current tasks via Thrimshim are:
+* `get_row`: Return a single row from the database by ID. Unauthenticated.
+* `update_row`: Updates row in the Events table. Authenticated and limited to Editors and Admins (who can update different sets of rows).
+* `manual_link`: Override the `video_link` field in the Events Table, in case of a manual upload. Authenticated and limited to Editors and Admins.
+* `reset_row`: Clear `state` and `video_link` columns and reset `state` to 'UNEDITED' in Events table. Authenticated and limited to Admins?
+
+Proposed actions:
+* `get_all_rows`: Return the entire events table (or specific subsets of it), for building dashboards. Unauthenticated
+* `submit_edits`: Rather than have have Thrimbletrimmer submit video edits to a generic update action/endpoint, have it go via a dedicated action that can only update the necessary actions. Authenticated to Editors and Admins.
+* `admin_update_row`: An "update row" action that can update all non-sheet-input fields as an "admin override".
 
 The planned roles are:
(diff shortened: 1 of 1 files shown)


ekimekim:
I don't think we need such a general system for defining roles. Editors should be able to do everything thrimshim currently exposes (effectively, their column list is "all" but they're still subject to the specific actions and checks coded), and admins should just have raw database access (the plan is to have some kind of off-the-shelf admin UI for them which basically lets you modify DB rows directly).
As for viewers, do we need to auth them at all? The only concern I would think of is gating access to who is allowed to restream as that could be a DoS vector, but that would mean authenticating in the restreamer which is a lot more annoying (since now backfiller has to auth against it, etc).

MasterGunner:
The granular permissions were because when I started thinking about it,
thrimshim only had the one generic endpoint/update action. If we're making
the endpoints more granular, then I agree we can do away with that
column... and really the Roles table entirely.

As for the viewer role - I just wanted to define it now in case we end up
deciding to lock things down more; since "viewer" is likely going to be a
"user role", if not a security role.

MasterGunner:
I updated the document to remove the "Role" table, and instead described the additional functions needed in Thrimshim to lock down access for who can update which columns.

SecurityModel.md
```diff
@@ -0,0 +1,42 @@
...
+The email address is then used to look up the user in the database to check their permissions.
+
+The current plan is to only authenticate/authorize users on datebase updates coming from Thrimbletrimmer; and since most use cases will 
+only have a single update event every few minutes, we can authenticate only as need - rather than authenticating on page load and storing session IDs.
+
+### Wubloader Authorization
+
+Current tasks via Thrimshim are:
+* `get_row`: Return a single row from the database by ID. Unauthenticated.
+* `update_row`: Updates row in the Events table. Authenticated and limited to Editors and Admins (who can update different sets of rows).
```
ekimekim: the existing thrimshim update_row call is basically the `submit_edits` call mentioned below, it can't do anything else.

SecurityModel.md
```diff
@@ -0,0 +1,42 @@
...
+
+Each user can only have one role.
+
+### Database Schema
+
+#### Members Table
+
+columns                    | type                               | description
+-------------------------- | ---------------------------------- | -----------
+`id`                       | `IDENTITY PRIMARY KEY`             | Unique account ID.
```
ekimekim: Given the limited usage of this table just for auth, I think we can keep it simple and just use email as the primary key.

ekimekim:
I'm wondering if we can get away with not having an admin role at all, and just defining admins as "people who have raw access to the database", ie. not by a webservice at all. If we've done our jobs right, admin intervention will only be required if things fuck up in an unexpected way.

I was expecting an admin web ui to be trivial to use an off-the-shelf solution for, but that appears to be false because the python-based ones all assume you're using an ORM, which we aren't.
Given how little we should need to use it, it's not worth spending any extra time on. So I think our admin interface should just be psql.

Then we can ditch role entirely, and just say "these endpoints require auth".

chrusher:
I agree with Ekim - the admins are going need raw access to the database anyway. Can we use pgAdmin to help with things?

MasterGunner:
If you're good with giving admins direct access to the database, that's fine with me.

MasterGunner:
Simplified the document down based on what we discussed. Even though it's pretty simple/straightforward now, still good to have written down.

MasterGunner:
It's something I'd like to add - hopefully along with a prototype of handling authentication - later this week (assuming I have the energy to do so).

chrusher:
If you wanted to update this to match the implementation, I'd be happy to approve this.

MasterGunner:
Should be up to date now, unless I forgot something.

PR #56, opened 2019-06-22
Chrusher/backfiller/database query

chrusher:
Getting a list of nodes to backfiller from the database

### DIFF
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -254,7 +254,7 @@ class BackfillerManager(object):
 		if self.node_database is not None:
 			self.logger.info('Fetching list of nodes from {}'.format(
 				urlparse.urlparse(self.node_database).hostname))
-			conn = self.db_manager.query()
+			conn = self.db_manager.get_conn()
 			results = database.query(conn, """
 				SELECT name, url, backfill_from
 				FROM nodes""")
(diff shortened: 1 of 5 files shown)


backfiller/backfiller/main.py
```diff
@@ -250,8 +252,15 @@ def get_nodes(self):
 						nodes[substrs[0]] = substrs[1]
 
 		if self.node_database is not None:
-			self.logger.info('Fetching list of nodes from {}'.format(self.node_database))
-			# TODO query the database
+			self.logger.info('Fetching list of nodes from {}'.format(
+				urlparse.urlparse(self.node_database).hostname))
+			conn = self.db_manager.get_conn()
```
ekimekim: This is going to make a new conn every time, which isn't ideal. The way most of our other components do this is that main() creates the dbmanager then passes the class a single conn to use.

backfiller/backfiller/main.py
```diff
@@ -250,8 +252,15 @@ def get_nodes(self):
...
+			results = database.query(conn, """
+				SELECT name, url, backfill_from
+				FROM nodes""")
```
ekimekim: Might as well filter server-side when possible: `SELECT name, url FROM nodes WHERE backfill_from`

PR #61, opened 2019-06-24
backfiller: Allow multiple concurrent segment downloads

ekimekim:
This will signifigantly increase throughput when downloading
large ranges of segments.

The max concurrency is exposed as a cli arg.

We also slightly modify the logged info, so it reports segments downloaded,
not just number of missing segments (which we might skip downloading for various reasons).

### DIFF
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -12,6 +12,7 @@ import uuid
 
 import argh
 import gevent.backdoor
+import gevent.pool
 import prometheus_client as prom
 import requests
 
@@ -142,7 +143,8 @@ class BackfillerManager(object):
 	NODE_INTERVAL = 300 #seconds between updating list of nodes
 
 	def __init__(self, base_dir, stream, variants, static_nodes=[], start=None,
-			run_once=False, node_file=None, node_database=None, localhost=None):
+			run_once=False, node_file=None, node_database=None, localhost=None,
+			download_concurrency=5):
 		"""Constructor for BackfillerManager.
 		Creates a manager for a given stream with specified variants. If
 		static_nodes is None, manager"""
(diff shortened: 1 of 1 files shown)


backfiller/backfiller/main.py
```diff
@@ -171,7 +174,7 @@ def start_worker(self, node):
 		if self.stopping.is_set():
 			logging.debug('Refusing to create new workers because we are stopping')
 			return
-		worker = BackfillerWorker(self, node)
+		worker = BackfillerWorker(self, node, self.download_concurrency)
```
chrusher: Minor point but the the workers get their other parameters from the manager rather than from the constructor arguments.

backfiller/backfiller/main.py
```diff
@@ -383,9 +400,11 @@ def run(self):
 @argh.arg('--node-file', help="Name of file listing nodes to backfill from. One node per line in the form NAME URI with whitespace only lines or lines starting with '#' ignored. If None (default) do not get nodes from a file.")
 @argh.arg('--node-database', help='Postgres conection string for database to fetch a list of nodes from. Either a space-separated list of key=value pairs, or a URI like: postgresql://USER:PASSWORD@HOST/DBNAME?KEY=VALUE . If None (default) do not get nodes from database.')
 @argh.arg('--localhost', help='Name of local machine. Used to prevent backfilling from itself. By default the result of socket.gethostname()')
+@argh.arg('--download-concurrency', help='Max number of concurrent segment downloads from a single node. Increasing this number may increase throughput but too high a value can overload the server or cause timeouts.')
```
chrusher: Should mention the default value in the doc string.

backfiller/backfiller/main.py
ekimekim: fixed

backfiller/backfiller/main.py
ekimekim: When the usage is printed, the default value is included automatically:
```
  --download-concurrency DOWNLOAD_CONCURRENCY
                        Max number of concurrent segment downloads from a
                        single node. Increasing this number may increase
                        throughput but too high a value can overload the
                        server or cause timeouts. (default: 5)
```

PR #64, opened 2019-06-29
Gunner/additional thrimbletrimmer integration

MasterGunner:
Added a "Get All Rows" function and REST call, a couple extra pieces of input validation, and a stub method for authentication later on.

### DIFF
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -48,8 +48,10 @@ def thrimshim(ident):
 		row = flask.request.json
 		return update_row(ident, row)
 	else:
-		return get_row(ident)
-		
+		if ident:
+			return get_row(ident)
+		else:
+			return get_all_rows()
 		
 def get_row(ident):
 	"""Gets the row from the database with id == ident."""
@@ -73,10 +75,39 @@ def get_row(ident):
 	}
 	return json.dumps(response)
 
+def get_all_rows():
+	"""Gets all rows from the events table from the database"""
+	conn = app.db_manager.get_conn()
+	results = database.query(conn, """
+		SELECT *
+		FROM events""")
+	
+	rows = []
+	row = results.fetchone()
+	while row:
+		row = row._asdict()
+		row['id'] = str(row['id'])
+		row = {
+			key: (
+				value.isoformat() if isinstance(value, datetime.datetime)
+				else value
+			) for key, value in row.items()
+		}
+		rows.append(row)
+		row = results.fetchone()
+	if not rows:
+		return 'No rows found', 404
+	
+	response = rows
+	return json.dumps(response)
+
 
 def update_row(ident, new_row):
 	"""Updates row of database with id = ident with the edit columns in
 	new_row."""
+
+	if not authenticate_user(new_row['Auth_Token']):
+		return 'User not authenticated.', 403
 
 	state_columns = ['state', 'uploader', 'error', 'video_link'] 
 	#these have to be set before a video can be set as 'EDITED'
@@ -93,6 +124,13 @@ def update_row(ident, new_row):
 	extras = set(new_row) - set(edit_columns + state_columns)
 	for extra in extras:
 		del new_row[extra]
+
+	#validate title length - YouTube titles are limited to 100 characters, 9 reserved for "DBXXXX - ".
+	if(len(new_row['video_title']) > 91):
+		return 'Title must be 91 characters or less', 400
+	#validate start time is less than end time
+	if(new_row['video_start'] > new_row['video_end']):
+		return 'Video Start must be less than Video End.', 400
 
 	conn = app.db_manager.get_conn()
 	#check a row with id = ident is in the database
@@ -140,6 +178,8 @@ def update_row(ident, new_row):
 	logging.info('Row {} updated to state {}'.format(ident, new_row['state']))
 	return ''
 
+def authenticate_user(authToken):
+	return True
 
 @app.route('/thrimshim/manual-link/<uuid:ident>', methods=['POST'])
 def manual_link(ident):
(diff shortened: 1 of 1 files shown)


thrimshim/thrimshim/main.py
```diff
@@ -40,6 +40,9 @@ def metrics():
 	"""Expose Prometheus metrics."""
 	return prometheus_client.generate_latest()
 
+@app.route('/thrimshim')
+def thrimshim_all_rows():
+	return get_all_rows()
```
ekimekim: Remove the needless indirection - just put `@app.route` on the `get_all_rows` function itself

thrimshim/thrimshim/main.py
```diff
@@ -73,11 +75,40 @@ def get_row(ident):
...
+def get_all_rows():
+	"""Gets all rows from the events table from the database"""
+	conn = app.db_manager.get_conn()
+	results = database.query(conn, """
+		SELECT *
+		FROM events""")
+	
+	rows = []
+	row = results.fetchone()
+	while row:
```
ekimekim: instead of this while loop, you can just iterate over the cursor:
```
for row in results:
```

thrimshim/thrimshim/main.py
```diff
@@ -73,11 +75,40 @@ def get_row(ident):
...
+		row = {
+			key: (
+				value.isoformat() if isinstance(value, datetime.datetime)
+				else value
+			) for key, value in row.items()
+		}
+		rows.append(row)
+		row = results.fetchone()
+	if not rows:
+		return 'No rows found', 404
```
ekimekim: If there's no rows, I'd rather return 200 and an empty list, ie. don't special case it.

thrimshim/thrimshim/main.py
```diff
@@ -73,11 +75,40 @@ def get_row(ident):
...
+	
+	response = rows
+	return json.dumps(response)
```
ekimekim: you can drop the variable rename, I think it's clear enough:
```
return json.dumps(rows)
```

thrimshim/thrimshim/main.py
```diff
@@ -94,6 +125,13 @@ def update_row(ident, new_row):
 	for extra in extras:
 		del new_row[extra]
 
+	#validate title length - YouTube titles are limited to 100 characters, 9 reserved for "DBXXXX - ".
+	if(len(new_row['video_title']) > 91):
```
ekimekim: style nitpick - spaces, no outermost parens
```suggestion
	if len(new_row['video_title']) > 91:
```

thrimshim/thrimshim/main.py
```diff
@@ -94,6 +125,13 @@ def update_row(ident, new_row):
 	for extra in extras:
 		del new_row[extra]
 
+	#validate title length - YouTube titles are limited to 100 characters, 9 reserved for "DBXXXX - ".
```
ekimekim: Hm. So I guess thrimbletrimmer is assuming that the cutter is automatically prepending that?
I'd rather avoid hard-coding those kinds of specifics in the cutter - maybe we should do that here instead? ie. write `DBXXXX - <user's input>` as the title value. Or more generally any arbitrary prefix.

On a related note, we also need to set tags, probably in the same place - either in the cutter, or here. Or maybe even in sheet sync. Let's discuss that further. This code is fine for now.

thrimshim/thrimshim/main.py
```diff
@@ -94,6 +125,13 @@ def update_row(ident, new_row):
...
+	if(len(new_row['video_title']) > 91):
+		return 'Title must be 91 characters or less', 400
+	#validate start time is less than end time
+	if(new_row['video_start'] > new_row['video_end']):
```
ekimekim: style nitpick - spaces, no outermost parens
```suggestion
	if new_row['video_start'] > new_row['video_end']:
```

thrimshim/thrimshim/main.py
```diff
@@ -73,11 +75,40 @@ def get_row(ident):
...
+	
+	response = rows
+	return json.dumps(response)
+
 
 def update_row(ident, new_row):
 	"""Updates row of database with id = ident with the edit columns in
 	new_row."""
 
+	if not authenticate_user(new_row['Auth_Token']):
```
ekimekim: I don't know if this is easier or harder on the client side, but generally I'd expect this kind of auth token to go in an `Authorization` header, not in the post data.
There's an advantage there in that most standard tools know not to log that field, treat it confidentially, etc.

thrimshim/thrimshim/main.py
MasterGunner: We had the cutter handle that in old Wubloader; so I'd just copied that forward. Makes perfect sense to just move the default title prefix into Thrimbletrimmer though; and keeps this section more generic. For now I'll switch to the 100 characters.

Previously all tags were set by the cutter automatically, since it was just a standard set of tags (plus the category). Maybe have Sheetsync set the default tags in an extra column, the Thrimbletrimmer can compile the final set (and possibly expose them to the editor), and submit them back?

thrimshim/thrimshim/main.py
MasterGunner: That comes from me not really knowing what I was doing, and me copying what I did last time (which was both a hackjob, and worked completely differently with authing on page load and generating session keys). Maybe just take this out for now?

ekimekim: Sure, we can address later.

thrimshim/thrimshim/main.py
ekimekim: Hm. Actually now I think about it, it probably makes sense to make both the title prefix (`DBXXXX - `) and tags be defined as part of the "upload location" config in the cutter, since it's youtube specific, and even more than that it's "main channel" specific. We don't want to do it for things we might want to upload to an alternate channel. Public vs unlisted falls under the same category.

PR #67, opened 2019-08-21
Chrusher/thrimshim/logging

chrusher:
Adding monitoring and more logging to thrimshim


common/common/stats.py
```diff
@@ -7,9 +7,11 @@
 import sys
 
 import gevent.lock
-
-import prometheus_client as prom
+from flask import request
+from flask import g as request_store
```
ekimekim: This introduces flask as a dependency to anyone that uses `common`, regardless of if they use flask.

Our approach so far for these kinds of things has been to put them in their own file in common (eg. database.py), and not import them in `__init__.py`, so that people have to explicitly import them to use them (and thus don't need their dependencies if they don't import them).

So I'd say let's take everything flask-specific and put it into a new `common/flask.py` file.

PR #72, opened 2019-08-30
Chrusher security database

chrusher:
This is a start at adding authentication to `thrimshim` by adding the `editors` table to the database. This also adds the `editor`, `edit_time` and `upload_time` to the `events` table and modifies `thrimshim` and `cutter` to update these columns at the appropriate times. Currently the `editor` field is set to `PLACEHOLDER` in `thrimshim` until the actually authentication code is written to provide the actual email.

### DIFF
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -41,12 +41,12 @@ def cors(app):
 
 
 
-def auth_wrapper(f):
+def authenticate(f):
 	"""Authenticate a token against the database.
 
 	Reference: https://developers.google.com/identity/sign-in/web/backend-auth"""
 	@wraps(f)
-	def decorated_function(*args, **kwargs):
+	def auth_wrapper(*args, **kwargs):
 		if app.no_authentication:
 			return f(*args, editor='NOT_AUTH', **kwargs)
 
@@ -76,11 +76,11 @@ def auth_wrapper(f):
 		return f(*args, editor=email, **kwargs)
 		
 
-	return decorated_function
+	return auth_wrapper
 
 @app.route('/thrimshim/auth-test', methods=['POST'])
 @request_stats
-@auth_wrapper
+@authenticate
 def test(editor=None):
 	return json.dumps(editor)
 
@@ -140,7 +140,7 @@ def get_row(ident):
 
 @app.route('/thrimshim/<uuid:ident>', methods=['POST'])
 @request_stats
-@auth_wrapper
+@authenticate
 def update_row(ident, editor=None):
 	new_row = flask.request.json
 	"""Updates row of database with id = ident with the edit columns in
@@ -219,7 +219,7 @@ def update_row(ident, editor=None):
 
 @app.route('/thrimshim/manual-link/<uuid:ident>', methods=['POST'])
 @request_stats
-@auth_wrapper
+@authenticate
 def manual_link(ident, editor=None):
 	"""Manually set a video_link if the state is 'UNEDITED' or 'DONE' and the 
 	upload_location is 'manual'."""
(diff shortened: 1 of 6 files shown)


chrusher:
Authentication tested with Thrimbletrimmer. Please review

thrimshim/thrimshim/main.py
```diff
@@ -36,6 +40,53 @@ def _start_response(status, headers, exc_info=None):
...
+	"""Authenticate a token against the database.
+
+	Reference: https://developers.google.com/identity/sign-in/web/backend-auth"""
+	@wraps(f)
+	def decorated_function(*args, **kwargs):
+		if flask.request.method == 'POST':
+			if app.no_authentication:
+				return f(*args, editor='NOT_AUTH', **kwargs)
+
+			userToken = flask.request.json['token']
```
ekimekim: A missing token should return a 402 (auth required), rather than erroring. A 400 is also acceptable (which might actually be what Flask does for you here. I know it does that for missing values in `request.params`, but I'm not sure about `request.json`)

thrimshim/thrimshim/main.py
```diff
@@ -15,6 +16,9 @@
 from common import database, PromLogCountsHandler, install_stacksampler
 from common.flask_stats import request_stats, after_request
 
+from google.oauth2 import id_token
+from google.auth.transport import requests
```
ekimekim: to avoid confusion with the `requests` library, and for clarity in general since these are super specific, please refer to them by their full names, ie `import google.oauth2.id_token` and `import google.auth.transport.requests` then use it as `google.oauth2.id_token.verify_oauth2_token(token, google.auth.transport.requests.Request(), None)`

thrimshim/thrimshim/main.py
```diff
@@ -36,6 +40,53 @@ def _start_response(status, headers, exc_info=None):
...
+			# check whether token is valid
+			try:
+				idinfo = id_token.verify_oauth2_token(userToken, requests.Request(), None)
+				if idinfo['iss'] not in ['accounts.google.com', 'https://accounts.google.com']:
+					raise ValueError('Wrong issuer.')
+			except ValueError:
+				return 'Invalid token. Access denied.', 403 
+
+			# check whether user is in the database
+			email = idinfo['email']
```
ekimekim: i'm a little worried about issues around multiple versions of the same email, eg. uppercase, or with extra dots in it. But I guess the version we get from google will always be the same, and we just need to ensure that the same version is the one in the database. Though a case-sensitive match would probably catch most of it.

thrimshim/thrimshim/main.py
```diff
@@ -186,25 +241,27 @@ def manual_link(ident):
 		return 'Row {} not found'.format(ident), 404
 	if old_row.state != 'UNEDITED' and not (old_row.state == 'DONE' and old_row.upload_location == 'manual'):
 		return 'Invalid state {} for manual video link'.format(old_row.state), 403		
-
+	now = datetime.datetime.utcnow()
 	results = database.query(conn, """
 		UPDATE events 
-		SET state='DONE', upload_location = 'manual', video_link = %s
+		SET state='DONE', upload_location = 'manual', video_link = %s,
+		editor = %s, edit_time = %s, upload_time = %s
```
ekimekim: nitpick: indent this second line of SET clauses, similar to line 264

thrimshim/thrimshim/main.py
chrusher: Do you mean case insensitive match?

ekimekim: yes, whoops

thrimshim/thrimshim/main.py
chrusher: 401 is Unauthorized. 402 is Payment Required.

ekimekim: right, derp.

chrusher:
Improvements based on ekim's suggests and a fix a bug discovered while retesting

thrimshim/thrimshim/main.py
```diff
@@ -41,51 +41,48 @@ def _start_response(status, headers, exc_info=None):
 
 
 
-def authenticate(f):
+def auth_wrapper(f):
```
ekimekim: I meant to rename `decorated_function`, not `authenticate` :P

PR #73, opened 2019-09-23
Chrusher database replication

chrusher:
Database replication and other database related issues.


INSTALL.md
```diff
@@ -54,5 +67,41 @@ To start the wubloader, simply run
 To stop the wubloader and clean up, simply run
 
   `docker-compose down`
-  
-To backfill from a node, the other nodes need to know about it. The best way to do this is to add the node to the database's nodes table.
+
+## Database setup
```
ekimekim: looks like this section is in here twice?

INSTALL.md
```diff
@@ -54,5 +67,41 @@ To start the wubloader, simply run
...
+* `db_args.user`, `db_args.password`, the username and password for the database user that the rest of the wubloader will connect to.
+* `db_super_user`, `super_password`, the username and password for the database superuser that is only accessible from the local machine. 
+* `db_replication_user`, `db_replication_password`, the username and password for the database user other nodes can connect as to replicate the database. If `db_replication_user` is an empty string, remote replication will be disabled.
+* `db_standby`, If true this database node will replicate the database node given by `db_args.host`. 
+
+It is recommended that the passwords be changed from the defaults in production.
+A database node needs to expose its database on a port. By default this is `5432` but the port exposed to the outside can be changed in the `ports` object.
+
+The `events` table will be automatically populated by the `sheetsync`. The startup script will attempt to populated the `nodes` and `editors` tables from the `nodes.csv` and `editors.csv` files in `segments_path` directory. The expected format for these files is:
```
ekimekim: typo
```suggestion
The `events` table will be automatically populated by the `sheetsync`. The startup script will attempt to populate the `nodes` and `editors` tables from the `nodes.csv` and `editors.csv` files in `segments_path` directory. The expected format for these files is:
```

ekimekim: What's the behaviour if these tables already exist and are populated, and the contents differ? Are they replaced? Merged (ie. anything new is added but otherwise unchanged)?

cutter/cutter/main.py
```diff
@@ -494,7 +494,16 @@ def main(dbconnect, youtube_creds_file, name=None, base_dir=".", metrics_port=80
 	# We have two independent jobs to do - to perform cut jobs (cutter),
 	# and to check the status of transcoding videos to see if they're done (transcode checker).
 	# We want to error if either errors, and shut down if either exits.
-	dbmanager = DBManager(dsn=dbconnect)
+	dbmanager = None
+	stopping = gevent.event.Event()
```
ekimekim: bad copy-paste, I think. We define `stop` as an event in this file already, you should use that.

db_scripts/db_dump
```diff
@@ -0,0 +1,13 @@
+#! /bin/bash
+NAME=${1:-postgres}
+CONTAINER=$(docker ps --format "{{.Names}}" | grep "$NAME")
+if [ -z "$CONTAINER" ]
+	then
+		echo "Container not found"
+		exit 1
+fi
```
ekimekim: We should probably also check that "$CONTAINER" contains exactly one result.
An easy way to do this would be to check that `[ "$(wc -l <<<"$CONTAINER")" -eq 1 ]`, ie. check there's exactly one line.

db_scripts/db_dump
```diff
@@ -0,0 +1,13 @@
...
+FILE="wubloader-$(date +%Y-%m-%dT%H:%M:%S).sql"
```
ekimekim: Rather than dumping to a temporary file then uploading the resulting file, we can skip the middleman and just pipe the result directly:
```
docker exec "$CONTAINER" pg_dump wubloader -U postgres | aws s3 cp - "s3://$BUCKET/$FILE"
```

db_scripts/db_dump
```diff
@@ -0,0 +1,13 @@
...
+echo "Dumping $CONTAINER to $FILE"
+docker exec $CONTAINER pg_dump wubloader -U postgres > $FILE
+aws s3 cp $FILE s3://wubloader-test
```
ekimekim: Bucket should be configurable.

db_scripts/db_restore
```diff
@@ -0,0 +1,25 @@
+#! /bin/bash
+
+if [ -z $1 ]
```
ekimekim:
```suggestion
if [ -z "$1" ]
```

db_scripts/db_restore
```diff
@@ -0,0 +1,25 @@
...
+	then
+		echo "No SQL script file"
```
ekimekim: Prefer outputting some useful usage info in this case, eg. `echo "USAGE: $0 SQL_SCRIPT"`

db_scripts/db_restore
```diff
@@ -0,0 +1,25 @@
...
+		exit 1
+fi
+
+NAME=${2:-postgres}
+CONTAINER=$(docker ps --format "{{.Names}}" | grep "$NAME")
+if [ -z "$CONTAINER" ]
+	then
+		echo "Container not found"
+		exit 1
+fi
```
ekimekim: as in dump, should ensure we've only found one matching container

db_scripts/db_dump
```diff
@@ -0,0 +1,13 @@
+#! /bin/bash
+NAME=${1:-postgres}
```
ekimekim: I think default search name should be `wubloader_postgres`. I think that's what docker-compose calls it.

db_scripts/db_restore
```diff
@@ -0,0 +1,25 @@
...
+
+#need to do this in case db dump file has a colon in it
+cp $1 tmp.sql
+docker cp tmp.sql "$CONTAINER:/"
```
ekimekim: rather than making a tempfile, you can pass it in via stdin:
```
docker cp - "$CONTAINER:/tmp.sql" <"$1"
```

db_scripts/db_restore
```diff
@@ -0,0 +1,25 @@
...
+rm tmp.sql
+
+docker exec $CONTAINER dropdb wubloader -U postgres --if-exists
+docker exec $CONTAINER createdb wubloader -U postgres
+docker exec $CONTAINER psql -d wubloader -f dump.sql -U postgres
+docker exec $CONTAINER rm tmp.sql
```
ekimekim: `$CONTAINER` needs quotes in all these.

docker-compose.jsonnet
```diff
@@ -53,21 +56,27 @@
   // You can exec into the container and telnet to this port to get a python shell.
   backdoor_port:: 1234,
 
-  // Other nodes to backfill from. You should not include the local node.
+  // Other nodes to always backfill from. You should not include the local node.
```
ekimekim: Probably add something like "If you're using the database to find peers, you should leave this empty"

postgres/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM postgres:latest
+COPY postgres/setup.sh /docker-entrypoint-initdb.d/setup.sh
+COPY postgres/standby_setup.sh /standby_setup.sh
+RUN chmod 0666 /docker-entrypoint-initdb.d/setup.sh
+RUN chmod 0700 /standby_setup.sh
```
ekimekim: nitpick: you should move the chmod for setup.sh to before the copy for standby_setup.sh, so that it doesn't need to be needlessly repeated if you're doing a local rebuild and only standby_setup.sh has changed.

postgres/setup.sh
```diff
@@ -0,0 +1,57 @@
+#! /bin/bash
+
+set -e
+
+sed -i "/host all all all/d" "$PGDATA/pg_hba.conf"
+echo "host all $WUBLOADER_USER all md5" >> "$PGDATA/pg_hba.conf"
```
ekimekim: add comment explaining what these two lines do? I can tell it's replacing a line in the config file, but what does that line mean?

docker-compose.jsonnet
```diff
@@ -53,21 +56,27 @@
...
-    "http://wubloader.codegunner.com/"
   ],
 
   // Connection args for the database.
   // If database is defined in this config, host and port should be postgres:5432.
   db_args:: {
-    user: "postgres",
-    password: "postgres",
+    user: "vst",
+    password: "dbfh2019", // don't use default in production
```
ekimekim: should note here that it must not contain `'` or `\`, since you're not doing any escaping in your setup script (nor do i think such escaping is worth bothering with)

this also applies to all the other password options.

postgres/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+
+if [ -a /mnt/wubloader/nodes.csv ]; then
+	echo "Loading nodes from nodes.csv"
+	psql -v -U postgres -d ${POSTGRES_DB} <<-EOF
+	CREATE TABLE IF NOT EXISTS nodes (
+		name TEXT PRIMARY KEY,
+		url TEXT NOT NULL,
+		backfill_from BOOLEAN NOT NULL DEFAULT TRUE);
+	COPY nodes FROM '/mnt/wubloader/nodes.csv' DELIMITER ',' CSV HEADER;
+	ALTER TABLE nodes OWNER TO vst;
```
ekimekim:
```suggestion
	ALTER TABLE nodes OWNER TO $WUBLOADER_USER;
```

postgres/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+fi
+
+if [ -a /mnt/wubloader/editors.csv ]; then
+	echo "Loading editors from editors.csv"
+	psql -v -U postgres -d ${POSTGRES_DB} <<-EOF
+	CREATE TABLE IF NOT EXISTS editors (
+		email TEXT PRIMARY KEY,
+		name TEXT NOT NULL);
+	COPY editors FROM '/mnt/wubloader/editors.csv' DELIMITER ',' CSV HEADER;
+	ALTER TABLE editors OWNER TO vst;
```
ekimekim:
```suggestion
	ALTER TABLE editors OWNER TO $WUBLOADER_USER;
```

postgres/setup.sh
```diff
@@ -0,0 +1,57 @@
...
+
+fi
+
+if [ -a /mnt/wubloader/nodes.csv ]; then
+	echo "Loading nodes from nodes.csv"
+	psql -v -U postgres -d ${POSTGRES_DB} <<-EOF
+	CREATE TABLE IF NOT EXISTS nodes (
+		name TEXT PRIMARY KEY,
+		url TEXT NOT NULL,
+		backfill_from BOOLEAN NOT NULL DEFAULT TRUE);
```
ekimekim: I'm concerned that these schemas are now written in two places. They could get out of sync, and there'd be no error.
Maybe it makes more sense to remove the schema stuff from common.database and put it entirely here?

postgres/standby_setup.sh
```diff
@@ -0,0 +1,21 @@
+#! /bin/bash
+
+
+if [ ! -s "$PGDATA/PG_VERSION" ]; then
+
+	pg_basebackup -d "host=$MASTER_NODE password=$REPLICATION_PASSWORD port=5432 user=$REPLICATION_USER" -D ${PGDATA} -vP
```
ekimekim: add comment explaining what this does?
also, why you're not setting `set -e` until AFTER it?

postgres/standby_setup.sh
```diff
@@ -0,0 +1,21 @@
...
+
+	set -e
+
+	cat > ${PGDATA}/recovery.conf <<-EOF
+	standby_mode = on
+	primary_conninfo = 'host=$MASTER_NODE password=$REPLICATION_PASSWORD port=5432 user=$REPLICATION_USER'
+	trigger_file = '/tmp/touch_to_promote_to_master'
```
ekimekim: should document this somewhere. "if the database fails, run this command to promote the replica"

postgres/standby_setup.sh
```diff
@@ -0,0 +1,21 @@
...
+	EOF
+
+	chown postgres. ${PGDATA} -R
+	chmod 700 ${PGDATA} -R
+
+fi
+
+gosu postgres postgres
```
ekimekim: what does this do?

sheetsync/sheetsync/main.py
```diff
@@ -280,7 +280,16 @@ def main(dbconnect, sheets_creds_file, edit_url, bustime_start, sheet_id, worksh
 
 	logging.info("Starting up")
 
-	dbmanager = DBManager(dsn=dbconnect)
+	dbmanager = None
+	stopping = gevent.event.Event()
```
ekimekim: again, `stop` already exists. use that.

thrimshim/thrimshim/main.py
```diff
@@ -219,15 +220,25 @@ def reset_row(ident):
 def main(connection_string, host='0.0.0.0', port=8004, backdoor_port=0):
 	"""Thrimshim service."""
 	server = WSGIServer((host, port), cors(app))
-	app.db_manager = database.DBManager(dsn=connection_string)
 
+	app.no_authentication = no_authentication
+	app.db_manager = None
+	stopping = gevent.event.Event()
```
ekimekim: `stopping` isn't being set anywhere, so it does nothing.
You should move the `def stop()` bit to before connecting to the DB, and add `stopping.set()` to the `stop()` function.

ekimekim:
This is great stuff. A couple of minor issues and a lot of nitpicking.

cutter/cutter/main.py
chrusher: I don't quite follow. `stop` is defined as a handler function not an event unless I am missing something about gevents (which is quite likely)

ekimekim:
```
stop = gevent.event.Event()
```
(line 489)

an Event is basically a boolean flag that can be wait()ed on. The `event.set()` function sets it to true and wakes up all waiters. On line 490 we're registering `stop.set()` to be called upon recieving a SIGTERM, and this will cause `stop.wait()` to immediately return.

postgres/setup.sh
chrusher: Agreed but I'll do the transfer after merging this to master.

cutter/cutter/main.py
chrusher: Ah, I see. I was confusing how the cutter and sheetsync do this with how the thrimshim does.

db_scripts/db_dump
chrusher: No, the default is directoryname_postgres_1.

db_scripts/db_restore
chrusher: You can only do that with tar files. Tarring and untarring would be more of a hassle that creating a tempfile.

chrusher:
The build failure is the lack of the `postgres` container in your quay account ekim.

PR #74, opened 2019-09-24
Chrusher database schema

chrusher:
Moved database schema out of `common/database.py` to `postgres/setup.sh`.

### DIFF
--- a/common/common/database.py
+++ b/common/common/database.py
@@ -13,71 +13,6 @@ import psycopg2.extras
 from psycogreen.gevent import patch_psycopg
 
 
-# Schema is applied on startup and should be idemponent,
-# and include any migrations potentially needed.
-SCHEMA = """
-
--- Create type if it doesn't already exist
-DO $$ BEGIN
-	CREATE TYPE event_state as ENUM (
-		'UNEDITED',
-		'EDITED',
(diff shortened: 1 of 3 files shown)


common/common/database.py
```diff
@@ -13,71 +13,6 @@
...
-CREATE TABLE IF NOT EXISTS editors (
-	email TEXT PRIMARY KEY,
-	name TEXT NOT NULL
-);
-
-"""
-
-
 class DBManager(object):
 	"""Patches psycopg2 before any connections are created, and applies the schema.
```
ekimekim: comment here needs updating

postgres/setup.sh
```diff
@@ -34,26 +34,78 @@ if [ -n "$REPLICATION_USER" ]; then
 
 fi
 
+echo "Applying schema for $POSTGRES_DB"
+psql -v ON_ERROR_STOP=1 -U $WUBLOADER_USER -d $POSTGRES_DB <<-EOSQL
+-- Create type if it doesn't already exist
+DO \$\$ BEGIN
```
ekimekim: since you're removing the "IF NOT EXISTS" from everything else, this DO...END block can also go away and you can just do the CREATE TYPE directly.

PR #75, opened 2019-09-24
Gunner/thrimbletrimmer integration

MasterGunner:
Merging in Thrimbletrimmer.

Change to Thrimshim: database query for all rows now ordered by start time.
Change to docker-compose: Added "thrimbletrimmer" as an option to enable, which will then set the location paths in nginx's config.
Changes to nginx's config: Included the mime.types file and gzip compression by default. If "thrimbletrimmer" is enabled, adds "/thrimbletrimmer" as a location (default location under that path is the editor page, "/" will also then 301 redirect to "/thrimbletrimmer/dashboard.aspx"). There are no separate metrics for the Thrimbletrimmer module, so that path is not created.

Other than those, just takes the Thrimbletrimmer module folder (all static content), and dumps it into nginx on image build.

History, and the source for the custom videojs-trimming-controls plugin (adds the trimming UI elements to the video.js library, which is used for cross-browser HLS support), can be found here: https://github.com/mastergunner/Thrimbletrimmer-IV


docker-compose.jsonnet
```diff
@@ -13,14 +13,15 @@
 
   // For each service, whether to deploy that service.
   enabled:: {
```
chrusher: For master we want to keep the default components - downloader, restreamer, backfiller, cutter, thrimshim, ngnix and now thrimbletrimmer enabled while disabling others.

docker-compose.jsonnet
```diff
@@ -224,6 +225,7 @@
     [if $.enabled.nginx then "nginx"]: {
       # mapping of services to internal ports for nginx to forward
       local forward_ports = {
+        thrimbletrimmer: 80,
```
ekimekim: thrimbletrimmer isn't really a seperate service (it's just an aspect of the nginx service), so it doesn't really have a sensical "port" (and indeed this value here is unused). I see two options:

1. Set port here to `0` and have a comment like `// thrimbletrimmer is served directly by nginx and its port value is unused`

2. Don't put thrimbletrimmer in the `SERVICES` env var at all. Remove it from this list and instead pass in a seperate env var `THRIMBLETRIMMER` = `"true"` or `""`. Then in the nginx config generate script, add it after line 21 as eg. `[ -n "$THRIMBLETRIMMER" ] && echo -e "\t\tlocation = / { return 301 \$scheme://\$host/thrimbletrimmer/dashboard.html; }\n\t\tlocation /thrimbletrimmer { }"`

I think option 2 is a lot cleaner, both in terms of the code here and in terms of the code in generate-config.
Thrimbletrimmer is a component but it isn't a service, as it does not have its own process or container. It's better to make it its own thing than to try to shoehorn it into being like a service but then add special cases to the concept of a service.

nginx/generate-config
```diff
@@ -14,8 +14,10 @@ LOCATIONS=$(
 		[ "$name" == "restreamer" ] && generate_location / "http://restreamer:$port"
 		# thrimshim takes any calls to thrimshim/
 		[ "$name" == "thrimshim" ] && generate_location /thrimshim "http://thrimshim:$port"
-		# all services have metrics under /metrics/SERVICE
-		generate_location "/metrics/$name" "http://$name:$port/metrics"
+		# thrimbletrimmer takes any calls to thrimbletrimmer/; serves content from /etc/nginx/html/thrimbletrimmer; requests to the plain URL get redirect to the dashboard page.
+		[ "$name" == "thrimbletrimmer" ] && echo -e "\t\tlocation = / { return 301 \$scheme://\$host/thrimbletrimmer/dashboard.html; }\n\t\tlocation /thrimbletrimmer { }"
+		# all services have metrics under /metrics/SERVICE, except for thrimebletrimmer
+		[ "$name" != "thrimbletrimmer" ] && generate_location "/metrics/$name" "http://$name:$port/metrics"
```
ekimekim: see my comment above, I think this can be done cleaner by not treating thrimbletrimmer as a service.

thrimbletrimmer/dashboard.html
```diff
@@ -0,0 +1,247 @@
...
+            let categoryfilters = Array.from(document.querySelectorAll("#CategoryFilter :checked")).map(checkbox => checkbox.getAttribute("filtervalue")).join();
+            let statefilters = Array.from(document.querySelectorAll("#StateFilter :checked")).map(checkbox => checkbox.getAttribute("filtervalue")).join();
+            window.location.href = `/thrimbletrimmer/dashboard.html?${startDateFilter ? "start="+startDateFilter+"&":""}${endDateFilter ? "end="+endDateFilter+"&":""}category=${categoryfilters}&state=${statefilters}`;
+        }
+
+        //On Page Load
+        getEvents().then(populateFilters).then(filterEvents).then(populateTable);
+    </script>
+</body>
+</html>
```
ekimekim: please ensure your files end in a newline, as per unix conventions. otherwise various bits of tooling may not handle it correctly.

thrimbletrimmer/plugins/video.js/dist/video-js.min.css
```diff
@@ -0,0 +1 @@
...set span{display:inline-block}.vjs-text-track-settings fieldset span>select{max-width:7.3em}.vjs-text-track-settings legend{color:#fff;margin:0 0 5px 0}.vjs-text-track-settings .vjs-label{position:absolute;clip:rect(1px 1px 1px 1px);clip:rect(1px,1px,1px,1px);display:block;margin:0 0 5px 0;padding:0;border:0;height:1px;width:1px;overflow:hidden}.vjs-track-settings-controls button:active,.vjs-track-settings-controls button:focus{outline-style:solid;outline-width:medium;background-image:linear-gradient(0deg,#fff 88%,#73859f 100%)}.vjs-track-settings-controls button:hover{color:rgba(43,51,63,.75)}.vjs-track-settings-controls button{background-color:#fff;background-image:linear-gradient(-180deg,#fff 88%,#73859f 100%);color:#2b333f;cursor:pointer;border-radius:2px}.vjs-track-settings-controls .vjs-default-button{margin-right:1em}@media print{.video-js>:not(.vjs-tech):not(.vjs-poster){visibility:hidden}}.vjs-resize-manager{position:absolute;top:0;left:0;width:100%;height:100%;border:none;z-index:-1000}.js-focus-visible .video-js :focus:not(.focus-visible){outline:0;background:0 0}.video-js .vjs-menu :focus:not(:focus-visible),.video-js :focus:not(:focus-visible){outline:0;background:0 0}
```
ekimekim: These files are obviously generated - generally the procedure here would be to have a way to build them instead of checking in the built result.

thrimbletrimmer/plugins/videojs-hls-quality-selector/dist/videojs-hls-quality-selector.min.js.bak
```diff
@@ -0,0 +1,7 @@
...
...;this.plugin.setQuality(this.item.value),this.selected(!0)},u}(i.getComponent("MenuItem")))(t,l,this._qualityButton,this)},n.prototype.onAddQualityLevel=function(){for(var n=this,t=this.player,o=t.qualityLevels().levels_||[],i=[],e=function(e){if(!i.filter(function(t){return t.item&&t.item.value===o[e].height}).length){var t=n.getQualityMenuItem.call(n,{label:o[e].height+"p",value:o[e].height});i.push(t)}},r=0;r<o.length;++r)e(r);i.sort(function(t,e){return"object"!==(void 0===t?"undefined":u(t))||"object"!==(void 0===e?"undefined":u(e))?-1:t.item.value<e.item.value?-1:t.item.value>e.item.value?1:0}),i.push(this.getQualityMenuItem.call(this,{label:t.localize("Auto"),value:"auto",selected:!0})),this._qualityButton&&(this._qualityButton.createItems=function(){return i},this._qualityButton.update())},n.prototype.setQuality=function(t){for(var e=this.player.qualityLevels(),n=0;n<e.length;++n){var o=e[n];o.enabled=o.height===t||"auto"===t}this._qualityButton.unpressButton()},n}(),e=function(n){var o=this;this.ready(function(){var t,e;t=o,e=i.mergeOptions(r,n),t.addClass("vjs-hls-quality-selector"),t.hlsQualitySelector=new l(t,e)})};return t("hlsQualitySelector",e),e.VERSION="0.0.8",e});
```
ekimekim: you appear to have accidentially checked in a `.bak` backup file.

thrimbletrimmer/scripts/IO.js
```diff
@@ -0,0 +1,168 @@
+//var desertBusStart = new Date("1970-01-01T00:00:00Z");
+var desertBusStart = new Date("2019-09-20T18:00:00");
```
ekimekim: Is there some way we can pass this in such that it can be defined in the main docker_compose.jsonnet config file?
It seems error prone to have multiple places where this value needs to be set, since they can get out of sync.

eg. I'd imagine this file gets minified at some point. Can we pass values into the minifier to set at "build time" to get baked into the minified file?

Alternately, we get `desertBusStart` from a thrimshim API call, or even just as part of the existing response objects.

docker-compose.jsonnet
MasterGunner: I did it this way just to get it working with minimal changes. I agree that it is a sufficiently different case from the other modules that it should be handled properly; but you know far better how to do it than I.

thrimbletrimmer/scripts/IO.js
MasterGunner: The 100% is not ideal, and I very much don't like it. It's just a value I needed to set somehow. 

Adding it in as a thrimshim API call would be the best way to do it right now; but preventing these kinds of extra round trips is also part of why I'd like to integrate thrimshim and thrimbletrimmer directly in a future iteration.

There's no build process at the moment; but that would certainly be another way to handle it.

Setting the stream Thrimbletrimmer is supposed to be asking for is another issue that faces the same problems.

ekimekim: Seems to me that having thrimshim return it as an extra field in its existing responses should be an easy solution without adding an extra round trip?

chrusher: Yup it can be easily added to `get_row()` in `thrimshim`. We can pass in the bus time as a command line option to `thrimshim` unless we want to add a bus_time_start table to the database. Adding it to the database is more work but that way we will get a consistent bus_time_start across all nodes.

docker-compose.jsonnet
MasterGunner: Fixed that.

chrusher:
What's the merge conflict?

ekimekim:
I'm not seeing any. Try refreshing.

chrusher:
Ah it's just a 'cannot be rebased' message. Sorry.

PR #79, opened 2019-09-28
Chrusher localhost fix

chrusher:
Fixes a bug in how the backfiller avoids backfilling from this node.

### DIFF
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -269,7 +269,7 @@ class BackfillerManager(object):
 				WHERE backfill_from""")
 			for row in results:
 				nodes[row.name] = row.url
-
+		nodes = {key:val for key, val in nodes.items() if val != self.localhost}
 		nodes.pop(self.localhost, None)
 		return nodes.values()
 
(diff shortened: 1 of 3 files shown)


backfiller/backfiller/main.py
```diff
@@ -269,7 +269,7 @@ def get_nodes(self):
 				WHERE backfill_from""")
 			for row in results:
 				nodes[row.name] = row.url
-
+		nodes = {key:val for key, val in nodes.items() if val != self.localhost}
```
ekimekim: so now we're both removing things where the url == self.localhost AND where the name == self.localhost. so is self.localhost meant to be a name or a url? I'm confused.

ekimekim: I think we should stick to node "names" as identifiers rather than URLs, especially since many different URLs can actually mean the same thing, eg. `http://example.com` vs `http://example.com/` vs `http://example.com:80/`

chrusher: Ok fair enough. I'll change the documentation to reflect this.

docker-compose.jsonnet
```diff
@@ -62,6 +62,7 @@
   peers:: [
   ],
 
+  localhost:: "http://example.com", // the name in the nodes table of the database
```
ekimekim: default value here should look like a node name, not a URL. eg. `mynode`

Issue #80, opened 2019-09-28
backfiller should verify segment hashes


ekimekim:
When backfilling, the backfiller should ensure the file it recieves matches the hash in the filename.

What to do when you have a mismatch is an open question. Let's think about what kinds of failures we expect to detect:
1. Corruption in transit - this is probably unlikely, and when it does happen it will almost always be due to the connection being closed early and the file truncated. In these cases it makes the most sense to just retry.
2. Corruption on disk on remote end - this is definitely a possibility, assuming we're not doing anything fancy like ZFS. In this case retrying won't help, and our only real remediation would be to delete the segment and let it replicate from another node. Of course, if that's the only copy, we have more issues. This should be a manual intervention, so the code should just error so we notice.
3. The file is intact but the hash is wrong - This is extremely unlikely. It would require either a particularly egregious logic error in the downloader, or a memory corruption bug (or disk corruption in the inode, but at that point you're probably going to corrupt the whole filesystem). Short of buying all nodes ECC RAM, there's not much we can do about it. This could also happen if someone is manually editing the segment filenames.

Basically, corruption in transit is the main thing to worry about and should be automatically retried. Beyond that, problems are probably hardware-related and require manual intervention. So I suggest we emit a warning, discard the recieved file, and just let it be retried next time we backfill. We should also add metrics for this failure condition.

chrusher:
I'll take a crack at this.

PR #93, opened 2019-10-06
Chrusher backfiller hash

chrusher:
Check the hashes of backfilled segments (as per #80). Also some more metrics.

### DIFF
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -2,6 +2,7 @@
 
 import datetime
 import errno
+import hashlib
 import logging
 import os
 import random
@@ -9,6 +10,7 @@ import signal
 import socket
 import urlparse
 import uuid
+from base64 import b64encode
 
 import argh
 import gevent.backdoor
@@ -88,6 +90,8 @@ def get_remote_segment(base_dir, node, channel, quality, hour, missing_segment,
 	temp_path = os.path.join(dir_name, "{}.ts".format(temp_name))
 	common.ensure_directory(temp_path)
 
+	hash = hashlib.sha256()
+
 	try:
 		logging.debug('Fetching segment {} from {}'.format(path, node))
 		uri = '{}/segments/{}/{}/{}/{}'.format(node, channel, quality, hour, missing_segment)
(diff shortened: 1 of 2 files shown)


backfiller/backfiller/main.py
```diff
@@ -26,6 +28,29 @@
...
+	['remote', 'channel', 'quality'],
+)
+
+hash_mismatches = prom.Counter(
+	'hash_mismatches',
+	'Number of segments with hash mismatches',
+	['remote', 'channel', 'quality', 'hour'],
+)
+
+get_node_errors = prom.Counter(
```
ekimekim: for consistency, can we call this variable `node_fetch_errors`?

ekimekim: actually, let's call it `node_list_errors`, and also change the line below. i think that's clearer as 'node fetch' can be confused with fetching segments from nodes

backfiller/backfiller/main.py
```diff
@@ -26,6 +28,29 @@
...
+	'node_fetch_errors',
+	'Number of errors fetching a list of nodes',
+	['filename', 'database'],
+)
+
+backfill_errors = prom.Counter(
+	'backfill_errors',
+	'Number of errors backfilling',
+	['remote', 'channel'],
```
ekimekim: no quality label?

backfiller/backfiller/main.py
```diff
@@ -210,6 +244,8 @@ def run(self):
 					failures += 1
 				delay = common.jitter(TIMEOUT * 2**failures)
 				self.logger.exception('Getting nodes failed. Retrying in {:.0f} s'.format(delay))
+				host = [s.split('=')[-1] for s in self.connection.dsn.split() if 'host' in s][0]
```
ekimekim: i'm worried about this line failing, especially since it only runs on error. maybe wrap in a try and set host='<unknown>' on error?

backfiller/backfiller/main.py
```diff
@@ -98,6 +124,14 @@ def get_remote_segment(base_dir, node, channel, quality, hour, missing_segment,
 		with open(temp_path, 'w') as f:
 			for chunk in resp.iter_content(8192):
 				f.write(chunk)
+				hash.update(chunk)
+
+		hash_str = b64encode(hash.digest(), "-_").rstrip("=")
+		if hash_str not in missing_segment:
```
ekimekim: I'm deeply suspicious of this line, but I can't come up with a scenario where it breaks for now.
What we should really be doing is passing in the parsed `segment` object to this function, instead of all the piecemeal components. then we can just compare to `segment.hash`.

backfiller/backfiller/main.py
chrusher: One backfill can do multiple quantities.

chrusher: I can easily move where the counter is so it counts each quality as its own backfill.

backfiller/backfiller/main.py
chrusher: Hopefully what I've just pushed does this in a more elegant and less error prone way.

chrusher:
Fixes #80

Issue #86, opened 2019-09-29
thrimbletrimmer issues and improvements


chrusher:
Since rows in the sheet get stored in the database once they have content and are not removed, any row that has text in it that subsequently gets removed will show up in the thrimbletrimmer dashboard as an empty event. Since rows get text added to them then removed all the time, the thrimblethrimmer dashboard should not display these events (or provide an option to display them).

It maybe useful to be able to sort columns in the dashboard.

It would likely be useful to further filter the unedited events in the dashboard to those marked for editing in the sheet. This would require marked for editing to be stored in the database.

How well does the dashboard handle large number of events? Last year there was about 800 rows in the sheet per day. By the end of the event this would mean the dashboard would be loading and displaying an potentially unwieldy 5000 events.

Does it make sense to have a button in thrimbletrimmer that saves the edit to the database but does not set the state to 'EDITED'?

chrusher:
For seeing videos stuck in the editing queue, it would be useful to see the editing_time. It would also be useful to reload the events from the database without reloading the page so you keep any filters or sorts in place.

MasterGunner:
Summarizing and adding comments:
- Be able to handle a line that was previously populated but later removed.
  - _Should be easy enough to filter out lines without any description. But maybe the database/sheetsync should have a column to indicate that an ID do longer exists/is populated in the database?_
- Sort columns in dashboard
  - _I'm not sure there's really a situation where we want to review events outside of time-order, but also an easy thing to add if you think there's a use case_
- Add filter for which lines are marked for editing.
  - _Absolutely would want to add this, but the database and sheetsync would need to be updated first._
- Handling of large numbers of events (5000+)
  - _Most clients *should* be able to handle the large number of events just fine; and it's easy enough to have a script populate the database until something breaks. However in an ideal world, we'd push the filtering back to the database (via Thrimshim), which is much more efficient. The database/thrimshim would also generate the lists of possible filters._
- A button that in thrimbletrimmer that saves the edit to the database but doesn't set the state to 'Edited'
  - _Like a "Save Draft" button? Probably would be useful; but we'll probably want an explicit "Draft" state column, if only to notify other editors - currently thrimshim would only accept "Edited" and "Unedited" back as State values._
- Add "editing_time" column to the dashboard
  - _Easy enough._
- Keep filters when reloading page/events
  - _Filters are stored in the query string of the URL, so they should be retained when you refresh the page (and your filters can be provided to others as a link)._
[...]

PR #94, opened 2019-10-07
Gunner/thrimbletrimmer improvements

MasterGunner:
Added a "Save Drafts" button to the editor page.
Added "Editing Time" and "Upload Time" to the dashboard page.
If a row doesn't have a description, it won't be displayed (to handle orphaned/cleared rows).

Currently live on editing.live for testing.

Handles the major requests from issue #86.


chrusher:
Looks good to me. Thanks for these changes Gunner.

PR #96, opened 2019-10-08
Chrusher cutter stats

chrusher:
Some stats on the number of videos cut and uploaded and associated errors

### DIFF
--- a/cutter/cutter/main.py
+++ b/cutter/cutter/main.py
@@ -373,7 +373,7 @@ class Cutter(object):
 					video_quality=job.video_quality,
 					upload_location=job.upload_location,
 					final_state='EDITED').inc()			
-			if not set_row(state='UNEDITED', error="Retryable error while uploading: {}".format(ex), uploader=None):
+			if not set_row(state='EDITED', error="Retryable error while uploading: {}".format(ex), uploader=None):
 				raise JobConsistencyError(
 					"No job with id {} and uploader {} when setting error while rolling back for retryable error"
 					.format(job.id, self.name)
--- a/cutter/cutter/main.py
+++ b/cutter/cutter/main.py
@@ -20,22 +20,11 @@ from common.segments import get_best_segments, cut_segments, ContainsHoles
 
 from .youtube import Youtube
 
-videos_cut = prom.Counter(
-	'videos_cut',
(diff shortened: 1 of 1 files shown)


cutter/cutter/main.py
```diff
@@ -20,6 +20,29 @@
...
+	'videos_cut',
+	'Number of videos successfully cut',
+	['video_channel', 'video_quality']
+)
+
+videos_uploaded  = prom.Counter(
+	'videos_uploaded',
+	'Number of videos successfully uploaded',
+	['video_channel', 'video_quality', 'upload_location']
+)
```
ekimekim: what's the difference between a video being cut and a video being uploaded? we do both at the same time.

ekimekim: reading further, it looks like the difference is in finalize beginning and finalize ending. maybe update the names to reflect that? say, `videos_finalizing` and `videos_uploaded`

chrusher: I think I will switch to just one counter for videos uploaded

cutter/cutter/main.py
```diff
@@ -315,11 +327,14 @@ def upload_wrapper():
...
 		except JobConsistencyError:
 			raise # this ensures it's not caught in the next except block
+			upload_errors.labels(video_channel=job.video_channel, video_quality=job.video_quality, upload_location=job.upload_location).inc()
 		except ErrorHandled:
 			# we're aborting the cut, error handling has already happened
+			upload_errors.labels(video_channel=job.video_channel, video_quality=job.video_quality, upload_location=job.upload_location).inc()
 			return
 		except Exception as ex:
 			self.refresh_conn()
+			upload_errors.labels(video_channel=job.video_channel, video_quality=job.video_quality, upload_location=job.upload_location).inc()
```
ekimekim: we should distinguish between errors in the cut vs errors in the upload, because cut errors are considered unretriable (final state is unedited) but upload errors are (final state is edited), under the assumption it's a transient network issue (though it could also be a youtube api issue which will fail every time).

though, we'd also want to distinguish between upload errors before / after finalizing, since ones after arent retriable and require operator intervention.

new plan: we could have one metric `upload_errors`, with a label for final state, ie. `unedited`, `edited` or `finalizing`.

PR #97, opened 2019-10-12
Chrusher sheetsync fixes

### LINKED ISSUE #90: Sheetsync needs to understand '--' in the end time column
Sheetsync needs to understand '--' in end time column as None not as a comment.

### LINKED ISSUE #92: Sheetsync does not pick up Tech Test/Preshow sheet
In my testing, I noticed the sheet sync doesn't pick up the Tech Test & Preshow sheet, since it doesn't fall in the Day 1-7 format.

I'm guessing the simplest way to do it would be to change the worksheets array in the docker compose to an arbitrary list of sheet names instead of a generated one in a specific format; or have two arrays (one generated, one arbitrary) that get joined together.

chrusher:
Sheetsync can now handle '--' in the time columns and negative times as well as parsing the Tech Test sheet. Also added a couple metrics for sheetsync.

Fixes #90 
Fixes #92

### DIFF
--- a/common/common/__init__.py
+++ b/common/common/__init__.py
@@ -27,7 +27,7 @@ def parse_bustime(bustime):
 	see it for detail."""
 	if bustime.startswith('-'):
 		# parse without the -, then negate it
-		return -parse_bustime(bustime[:1])
+		return -parse_bustime(bustime[1:])
 
 	parts = bustime.strip().split(':')
 	if len(parts) == 2:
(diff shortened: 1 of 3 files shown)


common/common/__init__.py
```diff
@@ -27,7 +27,7 @@ def parse_bustime(bustime):
 	see it for detail."""
 	if bustime.startswith('-'):
 		# parse without the -, then negate it
-		return -parse_bustime(bustime[:1])
+		return -parse_bustime(bustime[1:])
```
ekimekim: facepalm.jpg

PR #100, opened 2019-10-16
cutter: Allow multiple upload locations

### LINKED ISSUE #84: cutter should be able to upload to other locations
in addition to the default main yt channel, we should be able to configure:
* secondary YT channels (and whether uploads to this channel should be public or unlisted)
* twitch uploads (just in case)
* save to local disk (for things we can't upload and need to host ourselves. we could even set it up such that we save it directly into the place that such manual hosting is served from)

ekimekim:
This is a fairly large sequence of commits.
Consider looking at each individually as they approach this task in smaller,
more easily digestable steps.

We generalise the existing Youtube api object into a generic "upload backend".

We make this configurable, and set up a default config with two locations:
* A `desertbus` backend going to youtube
* An `unlisted` backend going to the same youtube, but with the videos unlisted.

We also add a 'local' upload backend that "uploads" to the local disk.
This will be useful for testing as well as self-hosting files later.

Fixes #84


cutter/cutter/upload_backends.py
```diff
@@ -0,0 +1,177 @@
...
+		filename = '{}-{}.ts'.format(title, video_id) # TODO with re-encoding, this ext must change
+		filepath = os.path.join(self.path, filename)
+		if self.write_info:
+			with open(os.path.join(self.path, '{}-{}.json'.format(title, video_id))) as f:
+				f.write(json.dumps({
+					'title': title,
+					'description': description,
+					'tags': tags,
+				}) + '\n')
+		with open(filepath, 'w') as f:
```
chrusher: Why all the different types for data?

PR #107, opened 2019-10-19
Chrusher segment coverage

### LINKED ISSUE #95: Metrics around segment coverage
This could be part of restreamer or its own service. It should periodically scan each hour and compile some metrics on how much of the hour is covered by segments, exporting them as gauges in seconds.

It should export:
* Total segment count and segment duration by segment type (full, partial, etc), ie. just sum all the durations without looking for overlap
* Raw coverage by best segment type. This should discard any overlap (eg. one segment from 00:10-00:12 and one segment from 00:11-00:13 would only count as 3 seconds of coverage). Coverage for non-"full" segment types should only cover parts where there's no full segment covering it, ie. the sum of raw coverage over all types should always be <= 1 hour.
* Editable coverage by best segment type. In contrast to raw coverage, editable coverage only counts segments that would be used if we were cutting a video. For example, if we had segments: `00:10-00:12, 00:11-00:13, 00:13-00:15`, then the raw coverage would be 5 seconds (union of all segment time ranges) but the editable coverage would only be 4 seconds (as `get_best_segments()` would first take the `00:10-00:12` segment, then it would discard the `00:11-00:13` segment because it overlaps the current segment, then it would have a 1sec hole before the `00:13-00:15` segment. This is unlikely to occur but if it happens we're going to be very glad we have metrics showing what's going on.
* Total overlapping segment count and overlap duration, by the lesser segment type of each pair (ie. so we can tell the difference between "we have both a full and a partial segment for this time" and "we have two conflicting full segments for this time").
* Count of holes, as determined by editable coverage and raw coverage each. Note we don't need to record total hole duration as that is implied by subtracting coverage from an hour.

chrusher:
Code to periodically generate segment stats and generate coverage images.

gamesdonequick_source_coverage
This is an example coverage image from the recent GDQx at TwitchCon.

Resolves #95

### DIFF
--- a/segment_coverage/segment_coverage/main.py
+++ b/segment_coverage/segment_coverage/main.py
@@ -118,25 +118,31 @@ class CoverageChecker(object):
 		self.stopping.set()
 
 	def create_coverage_map(self, quality, all_hour_holes, all_hour_partials,
-			pixel_length=2, hour_count=168):
-		"""Create a PNG show segment coverage.
+			pixel_length=2, hour_count=168, rows=300):
+		"""Create a PNG image showing segment coverage.
 
-		If any part of a pixel does not have coverage, it is marked as not
-		having coverage. Likewise, if only a partial segment is available for
-		any part of a pixel, it is marked as partial.
+		Each pixel repersents pixel_length seconds, with time increasing from
+		top to bottom along each column then right to left. By default each
+		pixel is 2 s and each column of the image repersents 10 min. White
+		pixels have no coverage, orange pixels only have coverage by partial
+		segments and blue pixels have coverage by full segments. If any part
+		of a pixel does not have coverage, it is marked as not having coverage.
+		Likewise, if only a partial segment is available for any part of a
+		pixel, it is marked as partial.
 
 		all_hour_holes -- a dict mapping hours to lists of holes
 		all_hour_holes -- a dict mapping hours to lists of partial segments
 		pixel_length -- length of a pixel in seconds
-		hour_count -- number of hours to create the map for"""
+		hour_count -- number of hours to create the map for
+		rows -- the height of the image"""
 
 		self.logger.info('Creating coverage map for {}'.format(quality))
 
 		latest_hour = datetime.datetime.strptime(max(all_hour_holes.keys()), HOUR_FMT)
 		hours = [latest_hour - datetime.timedelta(hours=i) for i in range(hour_count - 1, -1, -1)]
 
-		pixel_starts = np.arange(0, 3600, pixel_length) # starts of the pixels in 
-		pixel_ends = np.arange(pixel_length, 3601, pixel_length)
+		pixel_starts = np.arange(0, 3600, pixel_length) # start times of the pixels in an hour in seconds
+		pixel_ends = np.arange(pixel_length, 3601, pixel_length) # end times of the pixels in an hour in seconds
 
 		pixel_count = 3600 / pixel_length # number of pixels in an hour
 		coverage_mask = np.zeros(hour_count * pixel_count, dtype=np.bool_)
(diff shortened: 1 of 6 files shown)


segment_coverage/segment_coverage/main.py
```diff
@@ -0,0 +1,476 @@
...
+							if segment.type == 'full':
+								full_segments.append(segment)
+								full_segment_count += 1
+								full_segment_duration += segment.duration
+							elif segment.type == 'partial':
+								partial_segments.append(segment)
+								partial_segment_count += 1
+								partial_segment_duration += segment.duration
+						if full_segments:
+							if len(full_segments) == 1:
```
ekimekim: you can remove this if, because the code under the `else` will still do the right thing when length == 1. ie. it'll sort, pick the last (only) one, and then the for loop will loop 0 times since there's no other things in the list.

segment_coverage/segment_coverage/main.py
```diff
@@ -0,0 +1,476 @@
...
+								best_segment = full_segments[0]
+							else:
+								full_segments.sort(key=lambda segment: (segment.duration))
+								best_segment = full_segments[-1]
+								for segment in full_segments[:-1]:
+									full_overlaps += 1
+									full_overlap_duration += segment.duration
+							if partial_segments:
```
ekimekim: this `if` is superfluous, as looping over an empty list is already a no-op.

chrusher:
Using the GDQ segments, a normal hour is about 0.1 s on my desktop. Worst case (for an hour containing nearly 13 000 segments) it took 2 s. Generating the coverage map took less than 0.1 s.

In terms of differences with the behaviour of `segments.best_segments_by_start` and this, I believe the only difference is how duplicate partial segments are handled. `segments.best_segments_by_start` picks the segment with the largest file size while this picks the segment with the longest duration. I could change this behaviour if you want.

ekimekim:
yes, try to keep it consistent please.
the reasoning behind using size for partial instead of duration is because
duration is as reported in the playlist, not how much we actually got. we
approximate 'most whole' by largest.

segment_coverage/segment_coverage/main.py
```diff
@@ -0,0 +1,476 @@
...
+		self.logger = logging.getLogger('CoverageChecker({})'.format(channel))
+
+
+	def stop(self):
+		"""Stop checking coverage."""
+		self.logger.info('Stopping')
+		self.stopping.set()
+
+	def create_coverage_map(self, quality, all_hour_holes, all_hour_partials,
+			pixel_length=2, hour_count=168, rows=300):
```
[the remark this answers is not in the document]
chrusher: Should we provide defaults on the hour range and if so, what should they be?

ekimekim:
Earliest to latest hour is a reasonable default.

segment_coverage/segment_coverage/main.py
```diff
@@ -133,22 +109,32 @@ def create_coverage_map(self, quality, all_hour_holes, all_hour_partials,
...
+		if self.last_hour is None:
+			last_hour = datetime.datetime.strptime(max(all_hour_holes.keys()), HOUR_FMT)
+		else:
+			last_hour = self.last_hour.replace(minute=0, second=0, microsecond=0)
+		self.logger.info('Creating coverage map for {} from {} to {}'.format(quality,
+			first_hour.strftime(HOUR_FMT), last_hour.strftime(HOUR_FMT)))
+
+		hours = [first_hour]
+		latest_hour = first_hour
+		while latest_hour < last_hour:
```
ekimekim: This should be <=, or else hours won't include last_hour

chrusher: It does since the value of latest_hour that is appended to the hours list is one hour plus the value of latest_hour that is compared in the while statement.

ekimekim: ah, right. my mistake. maybe add comment explaining?

PR #109, opened 2019-10-20
warn about and ignore files that don't parse as segments

### LINKED ISSUE #98: best_segments_by_start should gracefully handle files that don't parse as segments
If common.segments.best_segments_by_start encounters a file it can't parse as a segment, it doesn't try to handle errors raised by parse_segment_path.

Rather than crashing whatever called it, best_segments_by_start should ignore files it can't parse as segments.

chrusher:
`common.segments.best_segments_by_start` now raises a warning and ignores for any files that it is unable to parse as segments rather than letting the `ValueError` from `parse_segment_path` pass.

Fixes #98

### DIFF
--- a/common/common/segments.py
+++ b/common/common/segments.py
@@ -11,6 +11,7 @@ import logging
 import os
 import shutil
 import sys
+import warnings
 from collections import namedtuple
 from contextlib import closing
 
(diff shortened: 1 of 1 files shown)


common/common/segments.py
```diff
@@ -11,6 +11,7 @@
 import os
 import shutil
 import sys
+import warnings
```
ekimekim: don't use `warnings`, no-one does outside the standard library. It only ever prints a given warning once, so it's extremely easy to miss. Use `logging.warning()`.

common/common/segments.py
```diff
@@ -222,7 +223,13 @@ def best_segments_by_start(hour):
...
 	# note we only parse them as we need them, which is unlikely to save us much time overall
 	# but is easy enough to do, so we might as well.
-	parsed = (parse_segment_path(os.path.join(hour, name)) for name in segment_paths)
+	# raise a warning for any files that don't parse as segments and ignore them
+	parsed = []
+	for name in segment_paths:
+		try:
+			parsed.append(parse_segment_path(os.path.join(hour, name)))
+		except ValueError as e:
+			warnings.warn(e)
```
ekimekim:
```suggestion
			logging.warning("Failed to parse segment {!r}".format(os.path.join(hour, name)), exc_info=True)
```

common/common/segments.py
```diff
@@ -222,7 +222,14 @@ def best_segments_by_start(hour):
 	segment_paths.sort()
 	# note we only parse them as we need them, which is unlikely to save us much time overall
 	# but is easy enough to do, so we might as well.
```
ekimekim: with this change, this comment is no longer true (you're parsing them all now, not lazily). I don't think the laziness was actually valuable, but we should delete the comment.

PR #119, opened 2019-10-27
Changes to improve behaviour if the DB is down

ekimekim:
For each component, we should do the best we can to continue to work at least somewhat
if the DB is unaccessible. See individual commits for discussion, but briefly:

* backfiller: Keep backfilling from the last known set of nodes
* thrimshim: Allow non-DB-related endpoints to work, 500 on DB-related ones
* cutter, sheetsync: Can't do anything useful without DB, just crash
* thrimbletrimmer: Allow non-row-specific editor window to work

(other components do not use the database and don't care if it's down)

### DIFF
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -189,9 +189,8 @@ class BackfillerManager(object):
 		self.start = start
 		self.run_once = run_once
 		self.node_file = node_file
-		self.node_database = node_database
-		if self.node_database is not None:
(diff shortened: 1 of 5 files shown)


ekimekim:
If the error is in connecting to the database, it would error inside the except block that catches errors from `get_nodes()`. Then it would exit due to the unhandled exception.

chrusher:
Good point. Just because I failed to break the previous version in my limited testing doesn't mean they can fail

PR #120, opened 2019-10-28
Chrusher backfiller delete

### LINKED ISSUE #110: Configurable window of "Last N Hours" of segments to keep
To minimize disk requirements on "just in case" capture nodes, we want to configure them to only keep, say, the last 24 hours of data.

This requires two features:
* Able to configure backfiller to only backfill up to N hours ago
* Some component (this could probably also be backfiller) to delete expired data.

chrusher:
Added the option to automatically delete hours older than the number of hours ago (or specific hour) given by `start`.

Fixes #110

### DIFF
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -6,7 +6,6 @@ import hashlib
 import logging
 import os
 import random
-import shutil
 import signal
 import socket
 import urlparse
@@ -161,6 +160,7 @@ def get_remote_segment(base_dir, node, channel, quality, hour, missing_segment,
 	segments_backfilled.labels(remote=node, channel=channel, quality=quality, hour=hour).inc()
 	logger.info('Segment {}/{}/{} backfilled'.format(quality, hour, missing_segment))
 
+
 def list_hours(node, channel, qualities, start=None):
 	"""Return a list of all available hours from a node.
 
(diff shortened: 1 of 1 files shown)


backfiller/backfiller/main.py
```diff
@@ -220,6 +242,53 @@ def stop_worker(self, node):
...
+			hours = [hour for hour in hours if datetime.datetime.strptime(hour, HOUR_FMT) < cutoff]
+			hours.sort()
+
+			for hour in hours:
+				path = os.path.join(self.base_dir, self.channel, quality, hour)
+				self.logger.info('Deleting {}'.format(path))
+				segments = list_local_segments(self.base_dir, self.channel, quality, hour)
+				for segment in segments:
+					try:
+						os.remove(os.path.join(path, segment))
```
ekimekim: we should have a metric tracking deleted segments, ideally with hour as label

backfiller/backfiller/main.py
```diff
@@ -262,6 +332,9 @@ def run(self):
 			if self.run_once:
 				break
 
+			if self.delete_old and self.start:
+				self.delete_hours()
```
ekimekim: if the DB is down / get_nodes() is erroring, deletes won't occur.
this isn't a problem, but it should be noted somewhere. even just a comment here is fine.

chrusher:
Given that deleting old hours isn't key to the operation of the backfiller, I don't want an error from deleting segments to crash the whole backfiller. Is it worth worrying about more exotic failure conditions than files/folders already being deleted or folders being non-empty?

ekimekim:
I think what you've got covers all expected conditions.

It might make sense to wrap the delete in a general catch and log warning about it, instead of crashing.

PR #128, opened 2019-10-30
Chrusher segement coverage fixes

### LINKED ISSUE #126: no/empty directories with segment_coverage
Segment coverage should not crash when the channel/quality directory does not exist nor should map creation crash everything if no hours are available.

### LINKED ISSUE #127: segment_coverage fails if the only coverage is temp segments
When the only segments available to segment_coverage for a start time are temp segments, segment_coverage fails

chrusher:
Fixes a couple bugs related to directories not existing or being empty as well as a more important one regarding temporary segments. 

Fixes #126 and fixes #127

### DIFF
--- a/segment_coverage/segment_coverage/main.py
+++ b/segment_coverage/segment_coverage/main.py
@@ -192,7 +192,7 @@ class CoverageChecker(object):
 					hours = [name for name in os.listdir(path) if not name.startswith('.')]
 				except OSError as e:
 					if e.errno == errno.ENOENT:
-						self.logger.warning('{} does not exist')
+						self.logger.warning('{} does not exist'.format(path))
 						break
 
 				hours.sort()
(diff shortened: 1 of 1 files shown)


segment_coverage/segment_coverage/main.py
```diff
@@ -183,7 +188,13 @@ def run(self):
 					break
 
 				path = os.path.join(self.base_dir, self.channel, quality)
-				hours = [name for name in os.listdir(path) if not name.startswith('.')]
+				try:
+					hours = [name for name in os.listdir(path) if not name.startswith('.')]
+				except OSError as e:
+					if e.errno == errno.ENOENT:
+						self.logger.warning('{} does not exist'.format(path))
+						break
```
ekimekim: I think this should be `continue` - even if one quality doesn't exist, others might. so we should try them all.

segment_coverage/segment_coverage/main.py
```diff
@@ -196,7 +207,12 @@ def run(self):
 					# based on common.segments.best_segments_by_start
 					# but more complicated to capture more detailed metrics
 					hour_path = os.path.join(self.base_dir, self.channel, quality, hour)
-					segment_names = [name for name in os.listdir(hour_path) if not name.startswith('.')]
+					try:
+						segment_names = [name for name in os.listdir(hour_path) if not name.startswith('.')]
+					except OSError as e:
+						if e.errno == errno.ENOENT:
+							self.logger.warning('{} does not exist'.format(hour_path))
```
ekimekim: slightly more descriptive error would be `Hour {} was deleted between finding it and processing it, ignoring`

PR #133, opened 2019-11-01
Add lots of metrics

ekimekim:
We have a generic metric wrapper for all our outgoing http calls.
We then use it in all services.

I also went over each service and added anything I thought was important.


ekimekim:
I thought about it - the problem is that the query is actually happening in the background - the returned cursor object will yield results as they arrive. So any latency measurements wouldn't make much sense, and calls could appear to succeed then still fail later.

chrusher:
Fair enough. We can view some metrics directly from the database using pgAdmin.

PR #134, opened 2019-11-01
Adds a dedicated streaming page for giffers etc, and more thrimbletrimmer things

### LINKED ISSUE #125: thrimbletrimmer: Should pad the given event start/end time
This has always been how thrimbletrimmer has worked, but it got lost in the new version of wubloader.

It's easy enough to re-add, just do it in the JS as the row is loaded.
I'm thinking -1min to start, and +2min to end (since the times are normally rounded down to the minute, so eg. if something actually ends at 1:10:59 it will show up as 1:10).

Note that with the new setup it's really easy to add more padding and re-load the playlist if needed, just by editing the start/end textboxes.

### LINKED ISSUE #130: thrimbletrimmer: Change bustime parsing to match elsewhere
In general, `XX:YY` bustime is parsed as `HH:MM`. (`XX:YY:ZZ` is always parsed as `HH:MM:SS`). But in thrimbletrimmer right now it's parsed as `MM:SS`. This should be fixed.

ekimekim:
Sorry, I thought I'd made a PR for this ages ago.

Fixes #130 
Fixes #125

### DIFF
--- a/restreamer/restreamer/main.py
+++ b/restreamer/restreamer/main.py
@@ -239,8 +239,16 @@ def cut(channel, quality):
 			A fast cut is much faster but minor artifacting may be present near the start and end.
 			A fast cut is encoded as MPEG-TS, a full as mp4.
 	"""
-	start = dateutil.parse_utc_only(request.args['start'])
-	end = dateutil.parse_utc_only(request.args['end'])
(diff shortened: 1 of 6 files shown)


ekimekim:
going commit by commit will give better context

PR #136, opened 2019-11-02
full cut: Optionally use seekable file OR directly stream

### LINKED ISSUE #135: cutter: Full cut: optionally allow streaming or buffering to file
Previously it would always stream the finished cut to the upload backend as it went. In 80d829b83be87a2c20bce6e52c578f74c8c80579 we changed it to always write to a seekable file on disk, then upload afterwards.

Ideally we want to be able to pick the correct behaviour based on the configured output format.

ekimekim:
The caller can pick depending on the needs of the output format.

This reverses most of 80d829b83be87a2c20bce6e52c578f74c8c80579, 
re-introducing streaming full cuts but keeping non-streaming as an option.

cutter,restreamer: Set stream=True for full cuts when appropriate

And also default to a new ffmpeg encoding setting for high-quality mpegts
(ie. still streamable) that is encoded very quickly.

Fixes #135

### DIFF
--- a/common/common/segments.py
+++ b/common/common/segments.py
@@ -295,8 +295,9 @@ def ffmpeg_cut_segment(segment, cut_start=None, cut_end=None):
 def ffmpeg_cut_stdin(output_file, cut_start, duration, encode_args):
 	"""Return a Popen object which is ffmpeg cutting from stdin.
 	This is used when doing a full cut.
-	Note the explicit output file object instead of using a pipe,
-	because most video formats require a seekable file.
+	If output_file is not subprocess.PIPE,
+	uses explicit output file object instead of using a pipe,
+	because some video formats require a seekable file.
 	"""
 	args = [
 		'ffmpeg',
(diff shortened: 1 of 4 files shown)


common/common/segments.py
```diff
@@ -400,40 +405,63 @@ def fast_cut_segments(segments, start, end):
...
-def full_cut_segments(segments, start, end, encode_args):
+def feed_input(segments, pipe):
+	"""Write each segment's data into the given pipe in order.
+	This is used to provide input to ffmpeg in a full cut."""
+	for segment in segments:
+		with open(segment.path) as f:
+			try:
+				shutil.copyfileobj(f, pipe)
+			except OSError as e:
+				# ignore EPIPE, as this just means the end cut meant we didn't need all i>
```
chrusher: Typo in the comment

ekimekim:
```suggestion
				# ignore EPIPE, as this just means the end cut meant we didn't need all it
```

PR #137, opened 2019-11-03
downloader: Create concept of an "important" channel

ekimekim:
In our usage, we have one channel where we really care / want to know if it's down,
but also a bunch of other channels where they're expected to not be streaming most/all of the time.

To prevent these extra channels making a ton of noise, we introduce the concept of an "important"
channel, indicated by appending a '!' to the channel name in the command line.

So for example, you might specify channels as "foo! foo_backup foo_behindthescenes".

Important channels have the same behaviour as previously.

Non-important channels:
* Have a 20-second retry on a master playlist fetch failure, instead of 5
* Log at debug when the stream is down, instead of info.

### DIFF
--- a/docker-compose.jsonnet
+++ b/docker-compose.jsonnet
@@ -25,7 +25,9 @@
   },
 
   // Twitch channels to capture. The first one will be used as the default channel in the editor.
-  channels:: ["desertbus"],
+  // Channels suffixed with a '!' are considered "important" and will be retried more aggressively
+  // and warned about if they're not currently streaming.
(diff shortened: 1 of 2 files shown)


chrusher:
Looks ok. Does the concept of important make sense for the backfiller or the segment_coverage components?

ekimekim:
I don't think it's worth it. This is a specific hack because for one channel we really care if the stream is down (becuase it should never be down) but for all the others we expect them to basically never be up.

PR #138, opened 2019-11-03
Allow thrimshim to safely cancel a job while it is cutting

### LINKED ISSUE #122: thrimbletrimmer: Add "Cancel Upload" button
that does the same thing as Reset Row, but only if the video has NOT yet been uploaded.
So if you accidentially submit then you can undo as long as the state is not yet FINALIZING,
and you can be sure there's no video to go clean up.

From thrimshim's perspective this should be a `force=false|true` flag that adds a `WHERE state = ...` and reports 403 if the state is wrong.

ekimekim:
This differs from the existing reset row by only suceeding if the upload is
not in finalizing.

We also make some changes to cutter to handle this situation gracefully.

We then add a button to thrimbletrimmer to do this under the name "Cancel Upload".

Fixes #122

### DIFF
--- a/cutter/cutter/main.py
+++ b/cutter/cutter/main.py
@@ -86,6 +86,10 @@ class CandidateGone(Exception):
 	"""Exception indicating a job candidate is no longer available"""
 
 
+class JobCancelled(Exception):
+	"""Exception indicating a job was cancelled by an operator after we had claimed it."""
+
+
 class Cutter(object):
 	NO_CANDIDATES_RETRY_INTERVAL = 1
 	ERROR_RETRY_INTERVAL = 5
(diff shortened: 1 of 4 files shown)


cutter/cutter/main.py
```diff
@@ -332,6 +339,10 @@ def set_row(**kwargs):
 			))
 			result = query(self.conn, built_query, id=job.id, name=self.name, **kwargs)
 			if result.rowcount != 1:
+				# If we hadn't yet set finalizing, then this means an operator cancelled the job
```
chrusher: To be pedantic, the lack of a row _likely_ means the job has been cancelled. The is still a very slim chance that a bug or database error could cause the row to go missing.

PR #140, opened 2019-11-03
Refactor dockerfiles for more shared layers

ekimekim:
By carefully ensuring most of our dockerfiles are identical in their first few layers,
we only need to build those layers once instead of every time.

In particular, we move installing gevent to before installing common,
so that even when common changes gevent doesn't need to be reinstalled.

This is important because gevent takes ages to install.

Also fixes segment_coverage, which wasn't being installed.


ekimekim:
I've piggybacked some other build improvements onto this PR too.

ekimekim:
Yeah, exactly. The time to install stuff like that is about the same as installing common + all its non-gevent dependencies, so it doesn't make a huge difference either way.

chrusher:
`readlink -f` breaks on OSX since `-f` isn't a valid option for the BSD `readlink` that OSX uses.

ekimekim:
ugh, of course. does it work if you replace it with `realpath`? (no -f)

chrusher:
realpath also errors on OSX but with either build or generate docker compose still works on OSX with a path that doesn't contain links.

PR #153, opened 2019-11-07
Add ability to explicitly pull and re-use layers from other commits when building

ekimekim:
This is intended mainly for travis CI, because by default it doesn't cache any layers
between builds.

By pulling likely-reusable builds (all parents of the current commit),
we take a fixed cost slowdown but in many cases should see a dramatic speed increase
overall, since we won't need to re-build anything that hasn't changed.

This isn't needed for local builds, where docker will do this on its own
with any previously-built images.


ekimekim:
This didn't help as much as I hoped, it looks like it's refusing to re-use some layers even when it should be able to. But it's still an improvement.

PR #157, opened 2019-11-09
added the ability to generate a webpage with all coverage maps

chrusher:


backfiller/backfiller/main.py
```diff
@@ -319,7 +319,7 @@ def run(self):
 				if failures < MAX_BACKOFF:
 					failures += 1
 				delay = common.jitter(TIMEOUT * 2**failures)
-				self.logger.exception('Getting nodes failed. Retrying in {:.0f} s'.format(delay))
+				self.logger.exception('Getting nodes failed. Retrying in {:.0f} s'.format(delay), exc_info=True)
```
ekimekim: This is redundant: logger.exception() is equivalent to logger.error(exc_info=True)

PR #173, opened 2020-09-22
Add alternate deploy stack for kubernetes

ekimekim:
This acts as an alternative to the docker compose setup.
It is intentionally nowhere near as fully featured and is currently only for replication nodes.

Notably, it uses k8s's Ingress mechanism to do routing to containers instead of our own nginx
(which is still present but just to serve static content).

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,4 +30,7 @@ A docker-compose file is provided to run all components. See `docker-compose.jso
 to set configuration options, then generate the compose file with `./generate-docker-compose`.
 Then run `docker-compose up`.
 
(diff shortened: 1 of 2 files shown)


chrusher (approved):
Looks OK

HubbeKing:
I can confirm it works as well. The downloader, restreamer, and segment_coverage containers have been healthy on my cluster now for ~48h. Haven't managed to test the backfiller yet, since I don't have peering info and the database isn't up yet.

HubbeKing:
Worth adding a snippet to the README.md about how it works, similar to the blurb about docker-compose?

PR #174, opened 2020-09-22
Add restreamer call to generate full concat'd videos

ekimekim:
This makes the timelapse easier.

We wrote this late last year and used it, but never got it merged

### DIFF
--- a/restreamer/restreamer/main.py
+++ b/restreamer/restreamer/main.py
@@ -292,6 +292,39 @@ def cut(channel, quality):
 		return "Unknown type {!r}".format(type), 400
 
 
+@app.route('/generate_videos/<channel>/<quality>.ts')
+@request_stats
+@has_path_args
+def generate_videos(channel, quality):
+	"""Generate one video for each contiguous range of segments (ie. split at holes),
+	and save them as CHANNEL_QUALITY_N.ts in the segments directory.
+	"""
+	start, end = time_range_for_quality(channel, quality)
+	hours_path = os.path.join(app.static_folder, channel, quality)
+	if not os.path.isdir(hours_path):
+		abort(404)
+
+	segments = get_best_segments(hours_path, start, end)
+	contiguous = []
+	n = [0]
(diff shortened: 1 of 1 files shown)


restreamer/restreamer/main.py
```diff
@@ -292,6 +292,39 @@ def cut(channel, quality):
...
+			return
+		with open(os.path.join(app.static_folder, "{}_{}_{}.ts".format(channel, quality, n[0])), 'w') as f:
+			for chunk in rough_cut_segments(contiguous, start, end):
+				f.write(chunk)
+		n[0] += 1
+
+	for segment in segments:
+		if segment is not None:
+			contiguous.append(segment)
+			continue
```
chrusher, line 322: It took me a long time to figure out what was going on with this loop.

Unless there is a subtly I am not getting I feel this could be written as `if segment is not None: contiguous.append(segment); else: write_file(); contiguous = []`

chrusher (commented):
Looks good except for some confusing flow of control.

restreamer/restreamer/main.py
ekimekim: yep, the two are equivalent. I'm used to using an "early return" style where I deal fully with one case and then return (or continue, like in this case) and then let the other case be un-indented. In something this short it doesn't really matter, but in a larger code block with more conditions, it can be the difference between this (imagine each `do_*` call was a larger code block):
```
for x in y:
  if foo:
    do_foo()
  else:
    do_thing()
    if bar:
      do_bar()
    else:
      do_baz()
```
and this:
```
for x in y:
  if foo:
    do_foo()
    continue
  do_thing()
  if bar:
    do_bar()
    continue
  do_baz()
```

We can use an `else` here if you prefer.

chrusher: I understand the idiom now.  It is fine the way it is.

chrusher (approved):
Looks good.

PR #175, opened 2020-09-22
sheetsync: Move column indexes to match updated sheet

ekimekim:
New tags column shunts all columns after it right by 1.
We will later want to parse that, but for now we ignore it.


chrusher (approved):
Looks good.

PR #177, opened 2020-09-25
Fix a bug when we fail to load a playlist that prevents loading other playlists

ekimekim:
This is a new failure case that occurs when the playlist call fails entirely (eg. 404
because the stream doesn't exist) instead of returning an empty playlist.


chrusher (approved):
Sure.

PR #178, opened 2020-09-26
Add TLS settings to k8s.jsonnet

HubbeKing:
I modified k8s.jsonnet to let me easily toggle ingress TLS on and off, figured I might as well get that change into the repo.

As it is, it simply adds a `tls` snippet to the Ingress spec if `$.config.ingress_tls` is true, and additionally adds a `secretName` to the `tls` snippet if `$.config.ingress_secret_name_required` is true. (getting the secretName from `$.config.ingress_secret_name`)

The added settings should handle most kubernetes ingress TLS situations, unless one's using cert-manager or something similar to auto-generate certificates and store those in ingress.spec.tls.secretName, that would require some metadata labels.
I use a wildcard cert myself so I haven't poked around with that yet.

Might add that change later, since I've been thinking of moving to per-subdomain certificates in my cluster to reduce reliance on my DNS provider.

### DIFF
--- a/k8s.jsonnet
+++ b/k8s.jsonnet
@@ -43,7 +43,14 @@
 
     // The hostname to use in the Ingress
     ingress_host: "wubloader.example.com",
-    
+
+    // Set to true to let the ingress handle TLS
+    ingress_tls: true,
+
+    // Set to true and give a secretName for ingress, if required for ingress TLS
+    ingress_secret_name_needed: false,
+    ingress_secret_name: "wubloader/tls",
+
     // Connection args for the database.
     // If database is defined in this config, host and port should be postgres:5432.
     db_args: {
@@ -232,8 +239,15 @@
             },
           },
         ],
+        [if ($.config.ingress_tls) then 'tls']: [
+            {
+                hosts: [
+                    $.config.ingress_host,
+                ],
+                [if ($.config.ingress_secret_name_needed) then 'secretName']: $.config.ingress_secret_name,
+            },
+        ],
       },
     },
   ],
-
 }
--- a/k8s.jsonnet
+++ b/k8s.jsonnet
@@ -47,9 +47,11 @@
     // Set to true to let the ingress handle TLS
     ingress_tls: true,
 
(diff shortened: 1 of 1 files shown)


HubbeKing:
Went ahead and added an `ingress_labels` setting and figured out how to check for field existance in jsonnet.

Should now be simple to use this to deploy to essentially any k8s cluster, no matter how TLS is handled in it.

k8s.jsonnet
```diff
@@ -201,7 +210,7 @@
       apiVersion: "networking.k8s.io/v1beta1",
       metadata: {
         name: "wubloader",
-        labels: {app: "wubloader"},
+        labels: {app: "wubloader"} + (if (std.objectHas($.config, "ingress_labels")) then $.config.ingress_labels else {}),
```
ekimekim, line 213: instead of checking for existence then defaulting to `{}`, use it here unconditionally then define it as `{}` instead of leaving it commented out (both work, but this is more consistent with our other config vars)

k8s.jsonnet
```diff
@@ -232,8 +241,15 @@
             },
           },
         ],
+        [if ($.config.ingress_tls) then 'tls']: [
```
ekimekim, line 244: `()` aren't required around if expressions
```suggestion
        [if $.config.ingress_tls then 'tls']: [
```

k8s.jsonnet
```diff
@@ -232,8 +241,15 @@
...
+            {
+                hosts: [
+                    $.config.ingress_host,
+                ],
+                [if (std.objectHas($.config, "ingress_secret_name")) then 'secretName']: $.config.ingress_secret_name,
```
ekimekim, line 249: `std.objectHas(x, y)` can be written more cleanly as `x in y`
```suggestion
                [if "ingress_secret_name" in $.config then 'secretName']: $.config.ingress_secret_name,
```

ekimekim (approved):
Looks good. Some style nitpicking.

ekimekim (approved):

PR #180, opened 2020-10-18
Add ability to set custom tags on videos

### LINKED ISSUE #169: Add support for custom tags
Currently the tags given a uploaded video are an event's category and sheet_name.

Going forward there will be a tag column in the sheet and tags should be editable in Thrimbletrimmer.

Non-exhaustive list of changes need to implement:

**postgres**

- Add `sheet_tags` and `edit_tags` columns to the event. We need two columns to keep things idempotent. 

**sheetsync**

- Read the tags in the tag column of the sheet into `sheet_tags`

**thrimshim**

- Handle updating the `edit_tags` column.
- Provide a list of existing tags.

**cutter**

- Include the `edit_tags` in the tags passed to the upload backends.

**Thrimbletrimmer**

- Display what tags are auto added to video
- Notify when a tag has not been used before

ekimekim:
We add two new database columns, `tags` as a sheet input, and `video_tags` as an edit input.

The editor can edit the tags. Category and sheet name are automatically included in the default
tags for the editor, but static tags are added at the cutter so the editor can't control them.

When cutting, we use `video_tags` as the list of tags to set on the video.

See commits for more details.

Fixes #169


sheetsync/sheetsync/main.py
```diff
@@ -224,6 +227,16 @@ def parse_row(self, row):
...
 					row_dict['_parse_errors'].append("Failed to parse column {}: {}".format(column, e))
 			row_dict[column] = value
+		# As a special case, add some implicit tags to the tags column.
+		# We prepend these to make it slightly more consistent for the editor,
+		# ie. it's always DAY, CATEGORY, POSTER_MOMENT, CUSTOM
+		row_dict['tags'] = (
+			[
+				row_dict['category'], # category name
+				worksheet, # sheet name
+			] + ('Poster Moment' if row_dict['poster_moment'] else [])
```
chrusher, line 237: It should be ['Poster Moment'] since you can't concat a string to a list

chrusher (changes requested):
Except for a tiny bug with poster moments this looks good.

sheetsync/sheetsync/main.py
ekimekim: whoops, thanks

chrusher (approved):
Looks good to me

PR #181, opened 2020-10-19
Add file and make as build dependencies

ekimekim:
gevent now requires these to build. I'm not sure when this changed.


chrusher (approved):
Looks ok to me

PR #182, opened 2020-10-19
Detect new style of twitch in-stream ads

### LINKED ISSUE #164: New twitch embedded ad form needs to be ignored
https://github.com/streamlink/streamlink/issues/2894#issuecomment-613692932

ekimekim:
New ad segments always have a title like "Amazon|...". This is the same fix
as used by streamlink at time of writing.

Fixes #164


chrusher (approved):
Looks ok to me

PR #183, opened 2020-10-19
Replace calls to gevent.signal() with gevent.signal_handler()

HubbeKing:
gevent.signal() was removed in gevent 1.5a4, see http://www.gevent.org/api/gevent.signal.html

Removed on Feb 5th, see https://github.com/gevent/gevent/pull/1530


chrusher (approved):
Looks ok to me

PR #184, opened 2020-10-20
Create new playlist manager service

### LINKED ISSUE #179: Playlist Manager
This year, youtube removed support for automatically populating playlists based on video tags.
We have many playlists and it is a chore to manage manually. So, we want to automate this.

This will take the form of a new component which watches for new videos in the DB, and adds them to playlists
based on their tags as listed in the DB. The mapping of playlist IDs to tags is static config. To avoid race conditions we should ensure only one playlist manager is running at once.
Note that we should do the right thing if we are restarted mid-run with a new playlist in the config, ie. it should find all previous videos and insert them then.

We need to be careful however, as adding to a playlist is a relatively expensive operation (50 quota, 1/32nd of uploading a video).
The good news is that operation includes inserting the video at any point in the playlist, so we can maintain correct order by inserting at the correct position every time. "Correct position" is done by event start time after mapping video ids back to events. We ignore unrecognized video ids (assumedly these were added manually) and insert after them on a tie (to prevent them getting "stuck" at the end of the playlist).

There is another quota issue. Listing playlist contents is meant to be cheap (1 quota), but each query only returns at most 50 items. We have some 1000-item playlists which can add up quickly.
We optimize to avoid unnessecary large reads by maintaining a mirrored copy of the playlist state locally. We validate this against the actual playlist by reading only one page and comparing the total playlist size (we assume manual changes will only ever be additions). This may turn out to be unnessecary if subsequent query pages do not count for quota - we need to test this.

A potential gotcha here: Previously we had several hard-coded tags added at upload time. For this to work properly all tags should be in the tags field. Default tags should instead be added at sheetsync. Make sure poster moment also creates an implicit tag.

ekimekim:
Playlist manager adds youtube videos to youtube playlists based on database
tags.

See #179 for full discussion.

Closes #179.

### DIFF
--- a/playlist_manager/playlist_manager/main.py
+++ b/playlist_manager/playlist_manager/main.py
@@ -1,0 +1,4 @@
+
+
+def main():
+	pass
--- a/playlist_manager/playlist_manager/main.py
+++ b/playlist_manager/playlist_manager/main.py
@@ -1,4 +1,308 @@
 
+import logging
+import json
+import signal
 
-def main():
-	pass
+import argh
+import gevent
+import gevent.event
+import prometheus_client as prom
+
+import common
+from common.database import DBManager, query
+from common.googleapis import GoogleAPIClient
+
+
+class PlaylistManager(object):
+
+	def __init__(self, dbmanager, api_client, upload_locations, playlist_tags):
+		self.dbmanager = dbmanager
+		self.api = YoutubeAPI(api_client)
+		self.upload_locations = upload_locations
+		self.playlist_tags = playlist_tags
+		self.reset()
+
+	def reset(self, playlist=None):
+		"""Called to clear saved state and force a refresh after errors.
+		Either reset a specific playlist, or all if no arg given.
+		"""
+		if playlist is None:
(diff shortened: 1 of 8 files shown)


playlist_manager/playlist_manager/main.py
```diff
@@ -0,0 +1,308 @@
...
+		return len(playlist)
+
+	def insert_into_playlist(self, playlist, video_id, index):
+		"""Insert video into given playlist at given index.
+		Makes the API call then also updates our mirrored copy.
+		"""
+		logging.info("Inserting {} at index {} of {}".format(video_id, index, playlist))
+		self.api.insert_into_playlist(playlist, video_id, index)
+		# Update our copy
+		self.playlist_state.setdefault(playlist, []).insert(0, video_id)
```
chrusher, line 164: Isn't this always adding the new video to the start of the local copy, not to the index it should be?

ekimekim: fixed

chrusher (approved):
Looks good now

ekimekim:
The build failure is unrelated - I'll make another PR to fix it.

PR #185, opened 2020-11-04
Add thrimshim to k8s.jsonnet

HubbeKing:
This PR adds thrimshim to the k8s.jsonnet file, so that k8s-based nodes can act as editors.

I need to fiddle with this some still and check that it works, though.

### DIFF
--- a/k8s.jsonnet
+++ b/k8s.jsonnet
@@ -18,7 +18,7 @@
     image_tag: "latest",
 
     // For each component, whether to deploy that component.
-    enabled:: {
+    enabled: {
       downloader: true,
       restreamer: true,
       backfiller: true,
--- a/k8s.jsonnet
+++ b/k8s.jsonnet
@@ -237,7 +237,7 @@
       $.db_connect,
       $.clean_channels[0],  // use first element as default channel
       $.config.bustime_start,
-      ]),
+    ]),
     // Normally nginx would be responsible for proxying requests to different services,
     // but in k8s we can use Ingress to do that. However nginx is still needed to serve
     // static content - segments as well as thrimbletrimmer.
--- a/k8s.jsonnet
+++ b/k8s.jsonnet
@@ -216,8 +216,9 @@
       if location != $.config.default_location]),
       $.db_connect,
       $.clean_channels[0],  // use first element as default channel
-      $.bustime_start,
+      $.config.bustime_start,
       ]
+    ),
     // Normally nginx would be responsible for proxying requests to different services,
     // but in k8s we can use Ingress to do that. However nginx is still needed to serve
     // static content - segments as well as thrimbletrimmer.
--- a/k8s.jsonnet
+++ b/k8s.jsonnet
@@ -1,8 +1,9 @@
 // This is a jsonnet file, it generates kubernetes manifests.
 // To generate and apply, run "jsonnet k8s.jsonnet | kubectl apply -f -"
 
-// Note this file is only set up to generate manifests for a basic replication node,
-// for the sake of simplicity.
+// Note that this file is currently not as advanced as its docker-compose variant
+// This file can only be used for replication nodes and editing nodes
+// see config.enabled for more info on what components can be used
 
 {
   kind: "List",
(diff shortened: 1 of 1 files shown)


k8s.jsonnet
```diff
@@ -158,23 +187,23 @@
   // (they may duplicate work, but not cause errors by stepping on each others' toes).
   items: [
     // The downloader watches the twitch stream and writes the HLS segments to disk
-    $.deployment("downloader", args=$.config.channels + [
+    if $.config.enabled["downloader"] then $.deployment("downloader", args=$.config.channels + [
```
ekimekim, line 190: Note that a jsonnet expression like `if COND then EXPR` without an else will resolve to `null` when COND is false.
That means your `items` list will end up with `null` items in it.

Two ways to work around this:
* do `if COND then [ITEM] else []` to produce an array of either 1-item or 0-item arrays, then flatten
* filter out the nulls later (eg. `[item for item in list if item != null]`)

IMO the latter is simpler and easier to read.

k8s.jsonnet
```diff
@@ -186,24 +215,39 @@
...
     ]),
+    // Thrimshim acts as an interface between the thrimbletrimmer editor and the database
+    // It is needed for thrimbletrimmer to be able to get unedited videos and submit edits
+    if $.config.enabled["thrimshim"] then $.deployment("thrimshim", args=[
+      "--backdoor-port", std.toString($.config.backdoor_port),
+      "--title-header", $.config.title_header,
+      "--description-footer", $.config.description_footer,
+      "--upload-locations", std.join(",", [$.config.default_location] + 
+      [location for location in std.objectFields($.config.cutter_config)
+      if location != $.config.default_location]),
```
ekimekim, line 231: nitpick: formatting
```suggestion
      "--upload-locations", std.join(",", [$.config.default_location] + [
        location for location in std.objectFields($.config.cutter_config)
        if location != $.config.default_location
      ]),
```

k8s.jsonnet
```diff
@@ -186,24 +215,39 @@
...
+      $.db_connect,
+      $.clean_channels[0],  // use first element as default channel
+      $.config.bustime_start,
+      ]
+    ),
```
ekimekim, line 236: nitpick: formatting
```suggestion
    ]),
```

k8s.jsonnet
```diff
@@ -16,6 +17,20 @@
     // you're actually running, and must manually re-pull to get an updated copy.
     image_tag: "latest",
 
+    // For each component, whether to deploy that component.
+    enabled:: {
```
ekimekim, line 21: style nitpick: the `::` isn't needed here.

In jsonnet, `key::` instead of `key:` denotes a "hidden" field. Hidden fields do not render in the output and do not show up when you do `std.objectKeys()` (there is an alternate method to get all keys including hidden ones).

Generally the idea is that you can use a hidden key for "private" fields or intermediate values you don't want to include in the output.

In our `docker-compose.jsonnet`, the config options are all at the top level. This requires them all to be hidden keys to avoid rendering in the final output. IMO this was a mistake, just one we haven't bothered to correct, which is why I opted for a hidden top-level `config` here. Since `config` is hidden, it won't be rendered at all so the keys inside it do not need to also be hidden. It does no harm but is redundant.

HubbeKing:
Should be ready to rebase&merge, the CI failure is due to the psycopg2 issue that ekim's PR covers.

k8s.jsonnet
```diff
@@ -186,24 +218,40 @@
...
+      "--title-header", $.config.title_header,
+      "--description-footer", $.config.description_footer,
+      "--upload-locations", std.join(",", [$.config.default_location] + [
+        location for location in std.objectFields($.config.cutter_config)
+        if location != $.config.default_location
+      ]),
+      $.db_connect,
+      $.clean_channels[0],  // use first element as default channel
+      $.config.bustime_start,
+      ]),
```
ekimekim, line 240: indent nitpick
```suggestion
    ]),
```

ekimekim (approved):

PR #186, opened 2020-11-04
Use psycopg2-binary instead of psycopg2

ekimekim:
It's identical except we download pre-built binaries.
This makes builds faster and requires less dependencies.

### DIFF
--- a/backfiller/setup.py
+++ b/backfiller/setup.py
@@ -8,7 +8,7 @@ setup(
 		"argh",
 		"gevent",
 		"psycogreen",
-		"psycopg2",
+		"psycopg2-binary",
 		"python-dateutil",
 		"requests",
 		"wubloader-common",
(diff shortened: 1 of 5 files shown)


chrusher (approved):
Looks like something we should thought to do earlier.

HubbeKing:
The travis build still failed when installing psycopg2 though... cache hecking things up again, probably?

ekimekim:
It's definitely trying to install psycopg2-binary, it's just...still installing it from source. Ugh, I hate pip so much.

ekimekim:
Apparently this is because psycopg2 doesn't have a wheel for musl-based distros, and we use alpine.
I'm going to back out of this change and just make the source install work.

ekimekim:
Closing in favor of https://github.com/ekimekim/wubloader/pull/187

PR #187, opened 2020-11-04
Add dependencies required to install psycopg2 from source

ekimekim:
We can't install the binaries as they don't support musl


HubbeKing:
PR looks good, but the CI failure is due to the quay.io push going `unauthorized: access to the requested resource is not authorized` on playlist_manager. The other images pushed fine, and everything built.

ekimekim:
Ah, I didn't set up the new image repo properly. Will fix and re-run.

PR #188, opened 2020-11-08
downloader: Fix a bug when we can't find a particular stream quality

ekimekim:
The intended behaviour was to log a warning message and retry next time,
but still allow workers to be started for any streams found.

However, due to a missing continue, we fall through to attempting to start a worker
for a non-existent quality which causes a KeyError when looking up
`self.latest_urls[quality]`. This exception means we don't run through the other qualities,
so we never start any other quality.


chrusher (approved):
Good catch

ekimekim:
The credit goes to Ricin, who it seems hit this in practice.

PR #190, opened 2020-11-10
Disable stacksampler by default

ekimekim:
It causes problems due to the sheer number of unique metrics emitted, which makes
the prometheus endpoint be very expensive / fail a lot.

The data is not useful enough to justify the cost.


chrusher (approved):
A sensible change

PR #191, opened 2020-11-10
Default to only backfilling things from this year

ekimekim:
By setting a 6-month limit


rasek-sls (approved):
Looks good to me!

chrusher (approved):
Looks good to me

HubbeKing:
> Looks good to me!

Oops, this was actually me. That'll teach me to not check which account container I'm on in Firefox...

HubbeKing (approved):

PR #192, opened 2020-11-10
docker-compose/k8s: Limit segment-coverage start/end times

ekimekim:
Instead of defaulting to oldest/newest.


chrusher (approved):
Looks good

PR #194, opened 2020-11-10
Fix metrics on k8s

ekimekim:
Fixes k8s.jsonnet to have metrics rules for all services.

As a hack to work around the fact ingress doesn't rewrite the path,
we make restreamer and thrimshim allow `/metrics/ANYTHING` in addition to `/metrics` endpoints.

### DIFF
--- a/k8s.jsonnet
+++ b/k8s.jsonnet
@@ -289,10 +289,11 @@
               },
               local metric_rule(name) = rule(name, "/metrics/%s" % name, "Exact"),
               paths: [
-                // Map /metrics/NAME to each service (except restreamer)
+                // Map /metrics/NAME to each service
                 metric_rule("downloader"),
                 metric_rule("backfiller"),
-                metric_rule("segment-coverage"),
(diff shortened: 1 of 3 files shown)


chrusher (approved):
Ok

HubbeKing (approved):
Looks good to me! I'll have a look at using separate ingress objects for metrics and non-metrics some time later, since this'll work fine.

HubbeKing (changes requested):
One thing! `metric_rule("segment_coverage")` creates a rule for `/metrics/segment_coverage` that points at a service `wubloader-segment_coverage`, but the service is actually called `wubloader-segment-coverage`

HubbeKing:
I guess kubectl doesn't do verification on ingress.backend.serviceName =/

HubbeKing (approved):
Should be good now, looks fine in the jsonnet output at least, and a kubectl dry-run throws no errors.

ekimekim:
ship it

PR #195, opened 2020-11-10
playlist_manager: Compare tags case-insensitively

ekimekim:
To avoid confusion between eg. "Interview" and "interview"


chrusher (approved):
Ok

PR #200, opened 2020-11-12
Run segment coverage less often, and make that configurable

ekimekim:
Run it every 5min instead of every min.
We don't need that fine granularity on this data and it takes a lot of CPU.


chrusher (approved):
Ok

HubbeKing:
Don't you also need to add an `@argh.arg` decorator for the check_interval argument to main?

ekimekim:
It's not strictly needed but I've added it in order to have help text.

PR #201, opened 2020-11-13
common: When parsing segment timestamps, use decimal instead of float

ekimekim:
Floating point error leads to 1us differences in parsed times,
which causes false positives in the overlapping segments check.

By using a Decimal, we get the exact digits from the filepath.


chrusher (approved):
Ah yes, the joys of floating point numbers

PR #202, opened 2020-11-13
segment parsing: Hand-roll microsecond parsing

ekimekim:
float() is inaccurate and Decimal() is very slow (~3x the cpu usage)
so instead we right-pad with 0s (eg. so 1.2345 -> 1.234500) then convert to int microsec directly.


chrusher (approved):
Clever

PR #203, opened 2020-11-13
Thrimshim sheet editing checking

chrusher:
More helpful UI if the database has changed plus ignore changes that are just white space or capitalisation


ekimekim (approved):

ekimekim:
Don't want to roll the wubloaders during the initial run madness, but will get to it in a bit

PR #204, opened 2020-11-13
thrimshim: Fix a bug preventing submissions

ekimekim:
When comparing old and new video tags, it errors because it's a list, not string.

We change it to apply the transforms to all tags in the list, and also ignore changes in list ordering.


chrusher:
Thanks for fixing this

Issue #171, opened 2020-04-19
Lower default volume level in Trimbletrimmer


chrusher:
By default the editor volume should start at 50%.

A stretch goal would be for Thrimbletrimmer to remember the user's volume level.

HubbeKing:
Should be easy enough to do with an onLoad() script in thrimbletrimmer's index.html.

Set onload for the EditorContainer div to a script that runs this.volume("0.5") on videojs("my-player").
See: https://github.com/videojs/video.js/issues/2310

Remembering the user's volume level would be trickier... I guess it could be done with like, a cookie? I am not much of a frontend guy.

PR #207, opened 2020-11-16
Set video.js volume to 50% by default

HubbeKing:
When the video.js player is ready, set its volume to 50%

I'm not sure this will work, so I'm gonna let the CI build and test locally on my node.


chrusher (approved):
Worth a try.

HubbeKing:
Okay, this does work. This will close #171, we can make a new issue if we want to implement remembering user "desired" volume.

PR #210, opened 2020-11-18
Some improvements to thrimbletrimmer

ekimekim:
* Alert on errors in more places, especially on submit

* Don't redirect to the dashboard on submit, instead just alert


HubbeKing (approved):
Looks good to me!

This looks like it might also fix a nitpick I had with cancelling out of the "Row changed, click OK to submit anyway" dialog - it used to pop up a new alert with the changes if you clicked Cancel.

The change on L293-L300 looks like it sorts that out.

PR #213, opened 2021-01-24
Finally get around to committing the generate_videos code

ekimekim:
We use this endpoint to generate the full video files for the archive torrent.
I wrote this almost immediately after the run but never got around to making a PR.


chrusher (approved):
Looks ok to me.

ekimekim:
I've been putting off merging this because the build is broken. But I've decided to tackle that seperately since I think it will go away as part of #221

PR #222, opened 2021-09-12
port to python3

### LINKED ISSUE #221: Port codebase to python 3
Unfortunately we've reached a point where staying on python 2 is no longer a workable option. Our builds are fragile and our libraries buggy because upstream doesn't even bother to test python 2 compatability anymore. Most libraries are frozen at old versions.

We will need to be careful with our 2 -> 3 port as we work with binary data fairly often. Thankfully most of this binary data such as filenames are in a context where we can assume ascii only, so converting them to unicode is fine - we don't need to support non-utf-8 filepaths. Segment data will be truly binary though, and working with binary files around `subprocess.Popen` may be painful.

It would be nice if we could approach this issue piecemeal, but due to the `common` library that's not really doable. We would need to maintain a cross-compatible common, or else two copies, in order to work with some components in py2 and some in py3. In practice, it will be simpler to do a hard switch then test and deal with breakage - thankfully we don't need to worry about uptime outside of the run, so we have plenty of time.

ekimekim:
some basic testing has been done in downloader and restreamer,
but mostly this is just eyeballing and grepping for required changes.

Fixes #221

### DIFF
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -9,7 +9,7 @@ import os
 import random
 import signal
 import socket
-import urlparse
+import urllib.parse
 import uuid
 
 import argh
@@ -364,7 +364,7 @@ class BackfillerManager(object):
 		localhost name (given by the --localhost argument) are ignored to try
 		to prevent this node from backfilling from itself."""
 
-		nodes = {urlparse.urlparse(node).hostname:node for node in self.static_nodes}
+		nodes = {urllib.parse.urlparse(node).hostname: node for node in self.static_nodes}
 		
 		if self.node_file is not None:
 			self.logger.info('Fetching list of nodes from {}'.format(self.node_file))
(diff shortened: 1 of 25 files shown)


segment_coverage/segment_coverage/main.py
```diff
@@ -232,7 +232,7 @@ def create_coverage_page(self, quality):
 		final_path = '{}_coverage.html'.format(path_prefix)
 		common.ensure_directory(temp_path)
 		with open(temp_path, 'w') as f:
-			f.write(html)
+			common.writeall(f.write, html)
```
chrusher, line 235: Do we need the write wrapper here when we are just writing text to a normal file?

backfiller/backfiller/main.py
```diff
@@ -231,7 +231,7 @@ def __init__(self, base_dir, channels, qualities, static_nodes=[],
 	def stop(self):
 		"""Shut down all workers and stop backfilling."""
 		self.logger.info('Stopping')
-		for node in self.workers.keys():
+		for node in list(self.workers.keys()):
```
chrusher, line 234: Do we really need a list here?

chrusher:
I have looked over your changes and I do not see any problems. I think the only way we are going to find any subtle bugs lurking in the code is by using the code so please merge when you are ready.

segment_coverage/segment_coverage/main.py
ekimekim: No, but I felt it was safer to just do it everywhere.

backfiller/backfiller/main.py
ekimekim: Yes, becasuse `self.workers` is concurrently modified while we're looping, so we need to make a copy of the list as it was at the start of the loop.

PR #229, opened 2021-10-10
Display a waveform graph in thrimbletrimmer

### LINKED ISSUE #216: Show audio levels in thrimbletrimmer
It would help with rapid editing if there was a visual indication of where silences and noise are in the video timeline.

ekimekim:
This is helpful for finding long silences, etc.
We add a new restreamer endpoint to serve the image. Images are generated on the fly,
it's pretty cheap.

Fixes #216


chrusher (approved):
Looks ok to me

PR #239, opened 2021-10-30
Thrimbletrimmer Fixes 2021 Roundup #2

ElementalAlchemist:
This is a simpler set of fixes for additional issues found by Sokar (except for some color fixes, which were me). These are almost all visual/usability fixes.

- Fix the Submit Anyway button that appears when sheet data is changed being completely broken (no caption, not functioning)
- Improve several colors for the darker theme (entirely ones I forgot to change initially)
- Fix the successful submission message not appearing, and update it for multiple ranges
- Add "play from end" for ranges, which works like "play from start" (which jumps to the start point of a range) but for the end point of the range
- Fix resizing of the video description `<textarea>`
- Indicate errors for title and description entry in real time (as opposed to finding it on submission, though users will also find it on submission)

### DIFF
--- a/thrimbletrimmer/edit.html
+++ b/thrimbletrimmer/edit.html
@@ -98,6 +98,11 @@
 					alt="Set range end point to current video time"
 					class="range-definition-set-end click"
 				/>
+				<img
+					src="images/play_to.png"
+					alt="Play from end point"
+					class="range-definition-play-end click"
+				/>
 				<div class="range-definition-icon-gap"></div>
 				<img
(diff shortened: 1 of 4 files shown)


ElementalAlchemist:
Although #171 is already closed for this, I chose to take a comment from that issue anyway (since the new codebase otherwise made no changes to that functionality) and added the suggestion that Thrimbletrimmer can remember the user's volume setting.

chrusher (approved):
Looks ok to me

PR #240, opened 2021-10-31
Improve WSGIServer graceful shutdown handling

### LINKED ISSUE #226: restreamer may fail to shut down if SIGTERM delivered during startup
This is a race between when we install the SIGTERM handler, and calling `server.serve_forever()`.
In this case `server.stop()` is called before `server.serve_forever()`, and this has the effect of
the `stop()` being ignored.

Interestingly thrimshim already has a workaround for this, but it's a bit hacky.
I think something better would be along the lines of (pseudocode):

```
stopping = gevent.Event()
on sigterm:
	stopping.set()
serve = gevent.spawn(server.serve_forever)
gevent.wait([serve, stopping], count=1) # wait for either
server.stop()
serve.get() # re-raise if serve_forever() errored
```

ekimekim:
Previously both restreamer and thrimshim had some complex logic for dealing with
graceful shutdown, in different ways, that was still prone to race conditions.

We replace this with a common method that does it properly.

Fixes #226

### DIFF
--- a/common/common/__init__.py
+++ b/common/common/__init__.py
@@ -2,8 +2,12 @@
 """A place for common utilities between wubloader components"""
 import datetime
 import errno
+import logging
 import os
 import random
+from signal import SIGTERM
+
+import gevent.event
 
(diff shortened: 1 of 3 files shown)


chrusher (approved):
Looks good to me.

PR #242, opened 2021-10-31
cutter: Fix errors due to unsupported cut type not being handled correctly

ekimekim:
Previously, determining the correct cut type and setting up the cut iterator
wasn't an operation we expected to be able to fail.
It happened outside all error handling blocks.

However, due to the multi-range work, this can now fail if we are requesting
an unsupported combination of transitions and cut type.

In order to correctly handle this like other cut errors, ie. set to UNEDITED
and set error column, we move this logic into the upload_wrapper right before we iterate
through the resulting cut.


chrusher (approved):
Looks good

PR #243, opened 2021-10-31
Thrimbletrimmer Fixes 2021 Roundup #3

ElementalAlchemist:
This is a quicker round fixing a couple things of varying intensity:

- The range endpoints updating when the video load time changes turned out to be massively broken (further described in the relevant commit message). This makes it (and the interaction between ranges and reloading the video in general) work properly.
- Validation on title/description for a video now also applies on page load.
- Since a workflow for time conversion was mentioned during editor training, I added a time conversion tool to the restreamer page. It's not part of the header times like in the old Thrimbletrimmer because I found that to be fairly annoying when quickly entering times, but there is now a tool that does it.

### DIFF
--- a/thrimbletrimmer/index.html
+++ b/thrimbletrimmer/index.html
@@ -81,5 +81,38 @@
 			</div>
 		</div>
 		<a href="#" id="download">Download Video</a>
+		<a href="#" id="time-converter-link">Convert Times</a>
+		<form id="time-converter" class="hidden">
+			<h2>Time Converter</h2>
+			<div id="time-converter-time-container">
+				<input class="time-converter-time" type="text" placeholder="Time to convert" />
+			</div>
+			<img
+				src="images/plus.png"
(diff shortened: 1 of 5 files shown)


ekimekim (approved):
I'd still much prefer the time ago to work in the restreamer like it used to - the primary use case for restreamer is to catch something I just missed and being able to say "give me 15min ago" is much nicer than needing to work out when that was in bustime first. But that calculation isn't THAT onerous ("well it's 95:30 now so I want 95:15") so this is fine.

ElementalAlchemist:
It also already defaults to "time ago" so I'm not sure what you'd be converting for that?
My primary use case is picking up swaths of time that happened at some point during the run, for which entering times and then remembering that I didn't switch it to bustime *first* and then having to go back and enter the times a second time is lightly frustrating. (This happens occasionally.) I suppose I could fix that with some light UI rearrangement on the time entry.

PR #246, opened 2021-11-06
Set ghcr.io username to user triggering the build

HubbeKing:
This should hopefully make pushes work in PRs? I'm grasping at straws at this point.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,7 +42,7 @@ jobs:
         uses: docker/login-action@v1
         with:
           registry: ghcr.io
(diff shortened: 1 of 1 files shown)


ekimekim (approved):
Let's give it a try

HubbeKing:
So this appears to work. I'm unsure how the permissions work themselves out, but I'm *reasonably* sure that once this is merged, anyone with write access to the repo can open a PR and the workflow will build&push the images. No idea what happens if a Random Nobody opens a PR.

The documentation doesn't really say anything...

PR #260, opened 2021-11-17
Add support for generating chapter markers in the video description

ElementalAlchemist:
As requested, this adds support in Thrimbletrimmer for generating chapter markers for the YouTube video description. Chapter timestamps should pad with the video load range the same way time range timestamps do. The end result is appended to the video description when submitting to Wubloader.


ekimekim:
You are amazing!

PR #291, opened 2022-04-16
cache results of common.segments.best_segments_by_start

### LINKED ISSUE #287: Optimize HLS playlist generation
The restreamer spends most of its time iterating through segments in `get_best_segments` to serve large time ranges. This operation is slow, always produces the same result given the same input, and the input changes rarely, making it a perfect opportunity to throw a cache in front of.

ekimekim:
The restreamer spends most of its time iterating through segments (parsing them, determining the best one for each start time)
to serve large time ranges. Since this only depends on the list of filenames read from disk,
we can cache it for a given hour as long as that list is identical.

This is a little trickier than it sounds because `best_segments_by_start` is an iterator
and in most cases it won't be fully consumed. So we introduce a `CachedIterator` abstraction
that will both remember the previously yielded values, and keep track of the live iterator
so it can be resumed again if a previous invocation only partially consumed it.

This also has the nice side effect of merging simultaneous operations - if two requests come in
for the same hour at the same time, they'll share one iterator and both consume the results
as they come in.

Fixes #287

### DIFF
--- a/common/common/cached_iterator.py
+++ b/common/common/cached_iterator.py
@@ -1,0 +1,44 @@
+
+import itertools
+
+import gevent.lock
+
+
+class CachedIterator():
+	"""Wraps an iterator. When you iterate over this, it pulls items from the wrapped iterator
+	as needed, but remembers each one. When you iterate over it again, it will re-serve the
+	yielded items in the same order, until it runs out, in which case it starts consuming
+	from the wrapped iterator again.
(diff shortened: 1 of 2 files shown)


ekimekim:
Needs testing. Will trivially conflict with #290.

ekimekim:
Basic tests seem to work.

PR #294, opened 2022-09-02
chat archiver

ekimekim:
The chat archiver saves chat logs in JSON "batch" files under /CHANNEL/chat/HOUR/.
It uses backfiller to sync batch files between machines, and has a "merge" process where two
batches for the same time range are turned into one. In this way machines converge on a single
canonical log, even if one of them misses something.

See the documentation in the PR for more.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -34,6 +34,7 @@ jobs:
           - postgres
           - segment_coverage
           - playlist_manager
+          - chat_archiver
     steps:
       - name: Check out repo
         uses: actions/checkout@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -38,6 +38,8 @@ jobs:
     steps:
       - name: Check out repo
         uses: actions/checkout@v2
+        with:
+          submodules: recursive
 
       - name: Log into ghcr.io
         uses: docker/login-action@v1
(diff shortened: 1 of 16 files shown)


HubbeKing (changes requested):
The code itself looks good, but in order for the Github Actions workflow to also build the chat_archiver image, it needs to be added to the build matrix component list in .github/workflows/main.yml

ekimekim:
@HubbeKing any idea why building the chat archiver is failing with `denied: permission_denied: write_package` on push? I've given github actions write permission to the repo:
screenshot

HubbeKing:
Since it's only chat_archiver that's failing, there's probably something fucky with specifically that package's permissions? Try comparing the settings of https://github.com/ekimekim/wubloader/pkgs/container/wubloader-chat_archiver to https://github.com/ekimekim/wubloader/pkgs/container/wubloader-cutter ? Might be something subtly different and/or weird

ekimekim:
That was definitely the issue before, but I fixed it and still got the same error. Maybe I need to force an entirely new build before the credentials will update?

ekimekim:
Nope, still broken.

HubbeKing:
I don't get it. According to the docs, the GITHUB_TOKEN should be granted the permissions noted in the workflow file, and it clearly is, since it's successfully pushing the other component images. Could be something with specifically PRs? https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/enabling-features-for-your-repository/managing-github-actions-settings-for-a-repository#preventing-github-actions-from-creating-or-approving-pull-requests

ekimekim:
I'm going to try merging to see if it works on master?

chrusher (commented):
A couple minor comments

PR #295, opened 2022-09-04
Management of unlisted videos

ekimekim:
This PR adds a new database column "public", which is false for unlisted videos.

A video's public status can be set by the editor in the advanced pane.
It defaults to true.
This option is modifyable for already-uploaded videos, so we can initially upload the video
as unlisted then set it public later.

Unlisted videos are displayed in the sheet as state "UNLISTED" instead of "DONE", but that
is not a real state in the backend, an unlisted video is a normal video with public = false.

We also ensure unlisted videos are not added to playlists.


ekimekim:
There was some talk of making the public/unlisted thing a radio button or dropdown in the UI, but I'm lazy and the checkbox works well enough.

thrimbletrimmer/edit.html
```diff
@@ -205,6 +205,11 @@ <h2>Keyboard Shortcuts</h2>
 					<input type="checkbox" id="advanced-submission-option-allow-holes" />
 				</div>
 
+				<div>
+					<label for="advanced-submission-option-public">Make public (not unlisted)</label>
+					<input type="checkbox" id="advanced-submission-option-public" />
+				</div>
+
```
HubbeKing, line 212: Might be better to have the logic reversed here? As in, a default-unchecked checkbox for "Unlisted".

HubbeKing (commented):
Looks good, though a minor nitpick

PR #298, opened 2022-10-06
thrimbletrimmer: Add link to download the current player frame as an image

### LINKED ISSUE #276: restreamer: Give me frame at timestamp as image file

ekimekim:
This link is updated whenever the player changes time.
We do this in common.js as it is present in both edit and stream.

The intent here is to provide a way for people making custom thumbnails or doing other image stuff
to easily grab a clean copy of a frame from the stream.

Fixes #276

How it looks in the UI:
screenshot
The resulting download:
2021-11-15T06-29-58 182

### DIFF
--- a/thrimbletrimmer/scripts/common.js
+++ b/thrimbletrimmer/scripts/common.js
@@ -419,3 +419,37 @@ function videoPlayerTimeFromVideoHumanTime(videoHumanTime) {
 
 	return hours * 3600 + minutes * 60 + seconds;
 }
+
+function getSegmentList() {
+	return globalPlayer.latencyController.levelDetails.fragments;
+}
+
+function dateTimeFromVideoPlayerTime(videoPlayerTime) {
+	const segmentList = getSegmentList();
+	let segmentStartTime;
+	let segmentStartISOTime;
+	for (const segment of segmentList) {
+		const segmentEndTime = segment.start + segment.duration;
+		if (videoPlayerTime >= segment.start && videoPlayerTime < segmentEndTime) {
+			segmentStartTime = segment.start;
+			segmentStartISOTime = segment.rawProgramDateTime;
+			break;
+		}
+	}
+	if (segmentStartISOTime === undefined) {
+		return null;
+	}
+	const wubloaderDateTime = DateTime.fromISO(segmentStartISOTime);
+	const offset = videoPlayerTime - segmentStartTime;
+	return wubloaderDateTime.plus({ seconds: offset });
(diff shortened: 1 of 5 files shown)


thrimbletrimmer/scripts/common.js
```diff
@@ -313,6 +313,27 @@ function setUpVideoControls() {
 			globalSeekTimer = null;
 		}
 	});
+
+	updateFrameLink(videoElement.currentTime);
+	videoElement.addEventListener("timeupdate", (_event) => {
+		updateFrameLink(videoElement.currentTime);
```
ElementalAlchemist, line 319: Can/should this be part of the existing `"timeupdate"` event (above)?

thrimbletrimmer/scripts/common.js
```diff
@@ -313,6 +313,27 @@ function setUpVideoControls() {
...
+}
+
+function updateFrameLink(videoTime) {
+	const frameLink = document.getElementById("download-frame");
+	if (frameLink !== undefined) {
+		const dateTime = dateTimeFromVideoPlayerTime(videoTime);
+		if (!dateTime) {
+			frameLink.href = "#";
+		} else {
+			const wubloaderTime = wubloaderTimeFromDateTime(dateTime)
```
ElementalAlchemist, line 330: A few things about this:
- It feels like doing a time conversion on every time update (which, granted, isn't every frame) might be approaching a lot. I don't want to convince the browser to give us fewer updates (assuming this has any influence).
- Following the above, note that the time might be inaccurate due to not getting a timeupdate for every frame. This could be up to a quarter second off (relevant for jump cuts, etc.).

Given that, it might be better if we can do this as a function that runs when the link is clicked to start the download of the relevant image (getting the current timestamp at that point, which we can query). [This thing I just found](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download) might be able to help a lot with that.

ElementalAlchemist (changes requested):

thrimbletrimmer/scripts/common.js
ekimekim: That seems reasonable, and makes more sense than this approach. I'll see if I can get it done while travelling.

ekimekim:
Updated with new approach, please take another look

ElementalAlchemist:
This is extremely a hack; good job.

Were there issues with the download API? Should I try to mess around with it later?

PR #299, opened 2022-10-10
Thrimbletrimmer support for thumbnails

ElementalAlchemist:
Adds Thrimbletrimmer support for all the new thumbnails fanciness. (This PR builds on #297 and therefore merges into that branch.)

This includes all the front-end UI for various useful tasks:
- Selecting the thumbnail type
- Choosing options based on which thumbnail type is specified
- Previewing the template (for the other three, we assume you know exactly what the thumbnail will look like or, for NONE, don't care)

Notable issues:
- Currently, custom thumbnails can't be uploaded. This seems to be a problem with thrimshim, but I'm not sure what it is. (It seems to fail on `b64decode`, which my install of Python does not for the same file/data, and it further fails trying to catch the error raised, so more useful information isn't shown to me.) I expect this should be resolved as part of #297 before it merges, but if there's secretly something I'm able to do about it, I'll give it another poke.

### DIFF
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -178,9 +178,9 @@ def get_row(ident):
 	]
 	if response['event_start'] is not None:
 		start = response['event_start']
-		if response['thumbnail_template'] is None:
+		if 'thumbnail_template' not in response or response['thumbnail_template'] is None:
 			response['thumbnail_template'] = DEFAULT_TEMPLATES[int(start.hour / 6)]
-		if response['thumbnail_time'] is None:
+		if 'thumbnail_time' not in response or response['thumbnail_time'] is None:
 			if response['event_end'] is not None:
 				# take full duration, and add half to start to get halfway
 				duration = response['event_end'] - start
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -178,11 +178,11 @@ def get_row(ident):
 	]
 	if response['event_start'] is not None:
 		start = response['event_start']
-		if 'thumbnail_template' not in response or response['thumbnail_template'] is None:
+		if response['thumbnail_template'] is None:
 			pst_hour = (start.hour - 8) % 24
 			shift = int(pst_hour / 6)
 			response['thumbnail_template'] = DEFAULT_TEMPLATES[shift]
-		if 'thumbnail_time' not in response or response['thumbnail_time'] is None:
+		if response['thumbnail_time'] is None:
 			if response['event_end'] is not None:
 				# take full duration, and add half to start to get halfway
 				duration = response['event_end'] - start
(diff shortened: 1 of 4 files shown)


thrimshim/thrimshim/main.py
```diff
@@ -178,11 +178,11 @@ def get_row(ident):
 	]
 	if response['event_start'] is not None:
 		start = response['event_start']
-		if response['thumbnail_template'] is None:
+		if 'thumbnail_template' not in response or response['thumbnail_template'] is None:
```
ekimekim, line 181: `'thumbnail_template' not in response` should never happen. we're fetching a database row, we can assume the schema is correct and all columns are present.

ElementalAlchemist: When I added this it's because it was crashing for not existing. It's *possible* I forgot to destroy the old database before running this (I don't remember now). I'll remove this change assuming that's the case.

ekimekim (approved):
This all looks great.
One note: Currently this code will always set `thumbnail_image = null` except when uploading a custom image. Ideally we would like to not modify `thumbnail_image` in a few scenarios:
* When updating the title/description of a video with a custom thumbnail (current implementation would need to re-upload the same custom thumbnail again in order to submit)
* If the thumbnail template has since changed, but we don't want to re-generate the thumbnails for existing videos. And we want to modify the title/description of such a video without affecting its thumbnail. This is admittedly pretty damn niche.

For the former, it would suffice to use the old `thumbnail_image` value originally loaded from thrimshim, unless a new one was uploaded.

For the latter, really we should omit `thumbnail_image` from the update params entirely if we don't want it to be updated. This would currently be rejected by thrimshim but I have a fix that will make optional modifiable columns work as intended. So it would be something like "only send `thumbnail_image = null` if `thumbnail_mode`, `thumbnail_time` or `thumbnail_template` differ from their original values".

To be honest we probably don't need the latter. But we should definitely have the former, so that we can safely tweak the title/description of videos with a custom thumbnail.

ElementalAlchemist:
Does it? The submit code does this:

```javascript
	if (thumbnailMode === "CUSTOM") {
		const fileInput = document.getElementById("video-info-thumbnail-custom");
		if (fileInput.files.length === 0) {
			if (!videoInfo.thumbnail_image) {
				submissionResponseElem.innerText =
					"A thumbnail file was not provided for the custom thumbnail";
				submissionResponseElem.classList.value = ["submission-response-error"];
				return;
			}
			thumbnailImage = videoInfo.thumbnail_image;
		} else {
			// snip (reads the file and base64s the data; see changes for full implementation)
		}
	}
```

I'm not sure how this doesn't handle the first case (please let me know what I'm missing). I'm happy to do the second as well, but it sounds like that should happen in tandem with the backend support for it.

ekimekim:
Oh, I missed the `if (!videoInfo.thumbnail_image)` part - my brain skipped straight to the error handling. That sounds good then.

ekimekim:
Manually merged after rebasing out the merge commits and handling conflicts.

PR #302, opened 2022-10-12
Add links to playlists in video descriptions

ekimekim:
For each video, we determine the set of playlists it should be in
and add a link to each one in the video description.

This applies when we update the row via thrimshim, so if a video's tags change
its description will not be updated until the video is modified in the editor.

To faciliate this we add two new columns to the tags sheet: a name, and whether or not
it should show in the description (eg. we don't want this for the All Everything playlist).

The description footer now looks like this:
```
This video is part of the following playlists:
- Hamfather [https://youtube.com/playlist?list=PL123456]
- Something Else [https://youtube.com/playlist?list=PLdeadbeef]
Uploaded by the Desert Bus Video Strike Team

### DIFF
--- a/sheetsync/sheetsync/main.py
+++ b/sheetsync/sheetsync/main.py
@@ -364,6 +364,11 @@ class SheetSync(object):
 		overwriting the entire playlists table"""
 		playlists = []
 		for row in rows:
+			# TODO redo sheet parsing for new columns
+			# Defaults for now:
+			name = ""
+			show_in_description = false
+
 			if len(row) != 3:
 				continue
 			tags, _, playlist_id = row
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by chrusher — COMMENTED** [review]
Looks ok to me

File: sheetsync/sheetsync/main.py [hunk 1]
```diff
@@ -364,24 +364,25 @@ def sync_playlists(self, rows):
 		overwriting the entire playlists table"""
 		playlists = []
 		for row in rows:
-			if len(row) != 3:
+			if len(row) != 5:
 				continue
-			tags, _, playlist_id = row
+			tags, _, name, playlist_id, show_in_desctiption = row
```
ElementalAlchemist, line 369: This typograghical error could be a problem later.

ElementalAlchemist:
If we're adding more stuff to the video description, we may want to adjust the validated maximum length of the video description in thrimbletrimmer/thrimshim downward (and maybe we should've done that for the "uploaded by VST" line too).

ekimekim:
We actually already do. The length check in line 308 occurs after all additions to the description. This does mean that if your description is over the max length with the additions but not without, then thrimbletrimmer won't catch it and instead you'll get a 400 from thrimshim. But I think that's acceptable for that kind of edge case, and we don't really have any reasonable way to predict how long the additions are going to be.

sheetsync/sheetsync/main.py
ekimekim: Fixed.

ElementalAlchemist:
That's what I meant. If we're fine calling it an edge case, we can leave it for now. I assumed a safety buffer of ~500 characters would cover most cases we'll run into.

PR #304, opened 2022-10-17
Show chat replay on restreamer page and transcript in editor

ElementalAlchemist:
Chat replay is here! This does all the things we've discussed in the meetings, plus other stuff I thought it'd be useful to support.

- The editor page shows a static transcript of all the chat that occurs during the loaded video.
- The restreamer page streams the chat as if chat is chatting during your view.
- In both cases, purged messages and users are shown lowlighted and crossed out. In the restreamer, the lowlight/crossout is applied at the time the CLEARCHAT/CLEARMSG is processed, which aligns with when it was sent in actual Twitch chat. (It was easier to implement this way).
- Emotes are supported.
- I added support for replies as well. If the message being replied to is in the chat replay on your page, you can click the reply to jump to the relevant message. (This was relatively easy to implement as a consequence of how I chose to implement CLEARMSG.)

I didn't end up referencing Didero's code basically at all--I looked at the start of it, saw how YouTube-specific the first lines of it were, and thought it'd probably be easier to just make my own thing than to convert that script's chat downloader to parse our JSON instead.

Note that chat is *not* supported for streaming (which is to say, you must have a start and end time specified to see chat--chat will not be loaded for unbounded streams). This would complicate my code a bit but also is not supported by the restreamer server. I think we can consider that for a standalone enhancement in the future. (I could consider an enhancement where we do the pagination ourselves in Thrimbletrimmer, but it's worth noting that the restreamer chat archive gets saved about once per minute, which makes accurately retrieving the chat JSON somewhat more difficult.)

This also pulls emotes directly from Twitch. When we do our emote archiving, we may wish to choose to swap emote fetching to be from Wubloader instead. I would propose that this be a wholly separate project/PR from this one.


thrimbletrimmer/scripts/common.js
```diff
@@ -444,22 +418,192 @@ function dateTimeFromVideoPlayerTime(videoPlayerTime) {
...
+	for (const chatLine of chatLogData) {
+		if (
+			chatLine.command !== "PRIVMSG" &&
+			chatLine.command !== "CLEARMSG" &&
+			chatLine.command !== "CLEARCHAT"
+		) {
+			continue;
+		}
+		const when = DateTime.fromSeconds(chatLine.time);
+		// Here, we just push each line successively into the list. This assumes data is provided to us in chronological order.
```
ekimekim, line 520: Some messages may be out of order as we aren't certain of their correct order, we only have a time range they may have occurred in (and they'll be sorted by start-of-time-range order).
However, PRIVMSG, CLEARMSG and CLEARCHAT will always have an exact timestamp and so always be in order.

thrimbletrimmer/scripts/common.js
```diff
@@ -444,22 +418,192 @@ function dateTimeFromVideoPlayerTime(videoPlayerTime) {
...
+	const response = await fetch(url);
+	if (!response.ok) {
+		return;
+	}
+	const chatLogData = await response.json();
+	for (const chatLine of chatLogData) {
+		if (
+			chatLine.command !== "PRIVMSG" &&
+			chatLine.command !== "CLEARMSG" &&
+			chatLine.command !== "CLEARCHAT"
```
ekimekim, line 515: You might also want to support USERNOTICE (displaying the text in `chatLine.tags["system-msg"]`). This includes sub messages (admittedly not a big deal for us) and raids.

thrimbletrimmer/scripts/stream.js
```diff
@@ -222,3 +228,94 @@ function convertEnteredTimes() {
...
+			if (globalChatData[lastChatIndex].when === lastAddedTime) {
+				break;
+			}
+			if (globalChatData[lastChatIndex].when < lastAddedTime) {
+				rangeMin = lastChatIndex;
+			} else {
+				rangeMax = lastChatIndex;
+			}
+			lastChatIndex = Math.floor((rangeMin + rangeMax) / 2);
+		}
```
ekimekim, line 297: binary search, nice :ok_hand:

ekimekim (approved):
Looks great, nicely done.

thrimbletrimmer/scripts/common.js
ElementalAlchemist: Ah, right. This didn't really come in my test data, but I'll see if I can add it at some point.

thrimbletrimmer/scripts/stream.js
ElementalAlchemist: yeah, I figured chat logs could be numerous (with DB chat being... DB chat, and particularly with long time ranges) and I wanted to be able to find the right chat quickly if possible.

thrimbletrimmer/scripts/common.js
ElementalAlchemist: In this case, I care specifically that `when` is in sequential order. If that's what matches with the unknown-here's-the-start-of-the-range timestamp (for other message types if I support them in the future), that's great.

PR #307, opened 2022-10-28
Add jump to chapter start and abbreviated YouTube titles in Thrimbletrimmer

ElementalAlchemist:
Here's the last couple nice-to-haves we have on the DB 2022 list. To make this work, I also fixed a bug that broke descriptions and reloading chapter marker data from videos that were already saved related to playlists in descriptions.


ElementalAlchemist:
We talked in IRC and found that the regex fix was actually for a completely different problem. Now that the completely different problem is fixed in master, that commit is removed.

ekimekim (approved):

PR #312, opened 2022-11-03
Don't show a template option for non-png files

ekimekim:
Each template now has two files, a `.png` and a `.json`. This is currently making them show up twice.
To fix this, we only consider files which end in `.png`.


ElementalAlchemist:
Should this be done like this, or should Thrimshim hide this implementation detail from the client?

ekimekim:
Yeah that's probably better. Will do.

ekimekim:
Done on the backend now.

ElementalAlchemist:
Awesome! Thanks!

PR #313, opened 2022-11-07
chat_archiver: Download each seen emote

ekimekim:
so we have a permanent record, in case they're deleted or changed later.

We'll continue to use twitch's copy in thrimbletrimmer as it will be faster (CDN).


chrusher:
Looks ok to me

ekimekim:
gah, forgot about this

PR #316, opened 2022-11-12
Thrimbletrimmer chat performance & updates

ElementalAlchemist:
This does a few things.

The thing I think will be most problematic that this fixes is the performance of chat updates in the restreamer. It turned out that, when seeking, the cost of calculating the video human time for each line of chat being rendered on screen was too high, causing the video to hang for several seconds while the chat code figured out all of the displayable timestamps. This calculation now occurs once on load in the background instead, so it's calculated for all messages before any chat data is displayed, and it does so without hanging the initial video load (nor, as before, does it cause the seek to hang). I expect this would've been a much bigger problem for restreamer than for the editor (which shows a static chat display that needs determined once on load and doesn't get updated/rerendered) but the cost was bad enough that I applied it to both. It likely would've taken a long time to load the editor for large videos otherwise.

I also made two other fixes, which I unfortunately couldn't really test (due to occurring outside of the stream being online):
- Clearing the whole chat, which I'm reasonably confident works
- Announcements, for which I expect we'll want to tweak the UI a bit (right now it works like all the other USERNOTICEs with "Announcement" in place of whatever system message was there before).

### DIFF
--- a/thrimbletrimmer/scripts/common.js
+++ b/thrimbletrimmer/scripts/common.js
@@ -604,7 +604,11 @@ function renderSystemMessages(chatMessageData) {
 	const systemTextElement = document.createElement("div");
 	systemTextElement.classList.add("chat-replay-message-text");
 	systemTextElement.classList.add("chat-replay-message-system");
-	systemTextElement.appendChild(document.createTextNode(chatMessage.tags["system-msg"]));
+	let systemMsg = chatMessage.tags["system-msg"];
+	if (!systemMsg && chatMessage.tags["msg-id"] === "announcement") {
+		systemMsg = "Announcement";
+	}
+	systemTextElement.appendChild(document.createTextNode(systemMsg));
 
 	const firstMessageContainer = createMessageContainer(chatMessageData, true);
 	firstMessageContainer.appendChild(sendTimeElement);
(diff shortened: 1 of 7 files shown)


thrimbletrimmer/scripts/common.js
```diff
@@ -579,12 +599,16 @@ function renderSystemMessages(chatMessageData) {
...
-	sendTimeElement.innerText = videoHumanTimeFromDateTime(chatMessageData.when);
+	sendTimeElement.innerText = chatMessageData.displayWhen;
 
 	const systemTextElement = document.createElement("div");
 	systemTextElement.classList.add("chat-replay-message-text");
 	systemTextElement.classList.add("chat-replay-message-system");
-	systemTextElement.appendChild(document.createTextNode(chatMessage.tags["system-msg"]));
+	let systemMsg = chatMessage.tags["system-msg"];
+	if (!systemMsg && chatMessage.tags["msg-id"] === "announcement") {
+		systemMsg = "Announcement";
```
ekimekim, line 609:
```suggestion
		systemMsg = `Announcement: ${chatMessage.params[1]}`;
```
This also shows the annoucement text.

ekimekim (approved):

thrimbletrimmer/scripts/common.js
ekimekim: Will add it myself post-merge.

ekimekim: oh, i see, you're adding it as a second message below. carry on.

ElementalAlchemist: yeah, that's already how the sub messages show, so I just kept it the same thing

PR #334, opened 2022-11-18
Properly handle when video or just metadata can be edited

ElementalAlchemist:
Note that I have no way to test this on my local Thrimbletrimmer, as I don't have an upload destination, so please try to give this more careful review.


ElementalAlchemist:
Update: Now that I've been informed how to set up local cutting, I think this should work. I'd still appreciate a decent double-check.

ekimekim (approved):

PR #340, opened 2023-06-26
Implement smart cuts

ekimekim:
Smart cuts re-write the timestamps in the mpeg-ts segments so that there are no weird jumps
and videos always start at time 0.


ekimekim:
I'm just going to merge this since we don't really have a proper system for getting review done.

PR #344, opened 2023-10-02
Backfill emotes directory

ekimekim:
We define a concept of "extras" which are non-segment directories that we want to backfill.
Right now the only "extras" directory is "emotes" but this might later include other things.

Restreamer has learned how to list these files, and backiller can backfill them.
Note that we still assume files with the same name have identical content
ie. if we already have a file, we don't re-download it.

### DIFF
--- a/backfiller/backfiller/main.py
+++ b/backfiller/backfiller/main.py
@@ -203,6 +203,33 @@ def list_hours(node, channel, quality, start=None):
 	return hours
 
 
+def list_local_extras(base_dir, dir):
+	root = os.path.join(base_dir, dir)
+	for path, subdirs, files in os.walk(root):
+		relpath = os.path.relpath(path, root)
+		for file in files:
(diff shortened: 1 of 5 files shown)


backfiller/backfiller/main.py
```diff
@@ -529,14 +558,45 @@ def backfill(self):
...
+				path = os.path.join(dir, file)
+				self.logger.info(f"Backfilling {path}")
+				workers.append(pool.spawn(
+					get_remote_extra(self.base_dir, self.node, path)
+				))
+
+		for worker in workers:
+			worker.get()
+
+		self.logger.info("Backfilled extras")
```
chrusher, line 588: I would include the number of files back filled to this info message - it is just len(missing_files)

chrusher (commented):
Looks good to me.

backfiller/backfiller/main.py
ekimekim: done

PR #349, opened 2023-10-29
Disallow angle brackets in chapter markers on the client

ElementalAlchemist:
Sokar wanted us to reject angle brackets in chapter titles... which we already do, but only on the server side. I chose to add error highlighting to the field when it happens so he stops complaining. :P

For the first commit, it looks like prettier had some sort of update. I ran it over all of thrimbletrimmer.

### DIFF
--- a/thrimbletrimmer/clock.html
+++ b/thrimbletrimmer/clock.html
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
 <html>
 	<head>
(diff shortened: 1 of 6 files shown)


ekimekim (approved):
Looks like pretter got trailing commas for function args enabled.
It's a bit weird for single-argument functions where it's only on its own line because the argument is long, but I don't really care either way.

Actual changes look good too.

PR #350, opened 2023-10-30
thrimbletrimmer: Change default download type to smart

ekimekim:
The new smart cut type avoids artifacting/playing issues that happen with fast cuts or multi-range rough cuts.

However it is new and experimental so we keep the old options available as backups.


chrusher (approved):
Looks good to me

PR #352, opened 2023-10-31
chat_archiver: Watch multiple channels at once

ekimekim:
Previously, the chat archiver could only watch a single channel.
This works ok for our purposes since there's really only one channel we care about,
but it means we can miss chatter on test channels or other weird situations.

This PR allows chat archiver to take multiple channels as arguments,
and connect to all of them with one connection. They are still batched seperately.

It also updates the docker-compose file to automatically watch all channels
we give to the downloader.


chrusher (approved):
Looks ok

PR #353, opened 2023-11-02
Get archive cuts working and integrated with the sheet

ekimekim:
This adds a `archive_worksheet` option to the docker-compose.
This specifies a worksheet where the torrent videos are defined.

We can then edit those videos the same way we do ones on the normal sheet.

They are treated differently in a few ways which required custom support:
* sheetsync: Different order of columns, with some not present
* thrimshim: Different default upload location, allow holes, no thumbnail, no title/description stuff.
* cutter: Uses an "archive cut" (merged already, but this fixes some bugs) that saves locally, with one file per continuous part of the stream,
  ie. it splits into a new file for every hole.


chrusher (commented):
Looks ok to me

PR #354, opened 2023-11-08
downloader: Add support for downloading from youtube, raw urls

ekimekim:
This adds support to downloader to read HLS streams from the following non-twitch sources:
* Youtube livestreams
* A raw HLS master playlist url

These alternate sources are limited in scope, eg. they don't support the full range of quality options.

Also, there is no guarentee we actually get MPEG-TS data from raw URL,
for example they may be DASH m4s files. We don't attempt to correct this in any way but save the segments
as given.

These changes are needed for some internal projects going on.


chrusher (approved):
A bit over my head but sure

PR #357, opened 2023-11-11
Hacky workaround for segments that can't be played independently

ekimekim:
This PR is just so images build


ekimekim:
Thankfully no longer needed

PR #362, opened 2023-11-14
Add 4x and 8x playback speeds

ElementalAlchemist:


ekimekim (approved):
Any chance you could make num + still go to 2x? Otherwise my muscle memory will kill me when skipping through.

ElementalAlchemist:
Should we say Shift+= is now 2x speed and not max speed? Or should we dedicate a new keyboard shortcut to it?

ekimekim:
Going to 2x would be my preference, since there's lots of levels between 1x and 2x but very few between 2x and higher. We could make the same button go to max if it's already >= 2x?

ElementalAlchemist:
I liked your last suggestion, so I have done that.

ekimekim (approved):

PR #367, opened 2023-11-15
Refresh just the images on the segment coverage page

ElementalAlchemist:
This allows users on the page to preserve attributes like the current horizontal scroll value.


chrusher (approved):
Looks good to me

ElementalAlchemist:
I shortened the query string variable name because why not.

PR #373, opened 2023-11-20
Update k8s.jsonnet file to support 2023 changes to images

HubbeKing:
These are the changes I made to make it actually function for the year's run, but I would like it looked over in case I did an Oopsie somewhere in it.

Could be that's why my node's metrics where a little Weird this year, after all.

### DIFF
--- a/k8s.jsonnet
+++ b/k8s.jsonnet
@@ -130,7 +130,7 @@
     downloader_creds_file: null,
 
     // The URL to write to the sheet for edit links, with {} being replaced by the id
-    edit_url: "https://wubloader.example.com/edit.html?id={}",
+    edit_url: "https://wubloader.example.com/thrimbletrimmer/edit.html?id={}",
 
     // The spreadsheet ID and worksheet names for sheetsync to act on
     sheet_id: "your_id_here",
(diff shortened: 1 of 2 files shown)


k8s.jsonnet
```diff
@@ -109,26 +113,26 @@
     // May be the same as cutter_creds_file.
     sheetsync_creds: import "./google_creds.json",
 
+    // Path to a file containing a twitch OAuth token to use when downloading streams.
+    // This is optional (null to omit) but may be helpful to bypass ads.
+    downloader_creds_file: null,
+
     // The URL to write to the sheet for edit links, with {} being replaced by the id
-    edit_url: "https://wubloader.example.com/thrimbletrimmer/edit.html?id={}",
+    edit_url: "http://thrimbletrimmer.codegunner.com/edit.html?id={}",
```
ekimekim, line 121: i've been trying to remove this old url from the codebase. the example.com one is fine.
```suggestion
    edit_url: "https://wubloader.example.com/thrimbletrimmer/edit.html?id={}",
```

ekimekim (approved):
all looks reasonable to me

k8s.jsonnet
HubbeKing: I have done this.

HubbeKing:
Using some more better mount options does appear to have sped things up considerably, but I suspect we won't know for sure without a proper test with backfill going and stuff.

PR #374, opened 2023-11-21
Allow a fudge factor when checking for gaps/overlaps between segments

ekimekim:
Sometimes in the wild (particularly on youtube) segments may not be timed
perfectly, so allow up to 10ms of gap or overlap to be counted as "equal"
for purposes of finding the best segment.

Also commit some segment coverage features I added while testing this.


chrusher (approved):
Seems quite reasonable

PR #375, opened 2023-11-26
Multiple improvements to performance and responsiveness of restreamer

ekimekim:
when requesting the same very large playlist over and over.

Main highlights:
* generating the hls playlist is now a streaming result
* if multiple calls for the same time range come in concurrently, we only run it once
* we yield to the event loop in more places to avoid locking up other requests


ekimekim:
Doing clean up of open PRs. I did some self-review, and I figure that "me, 5 months later" is sufficiently removed to count as a second set of eyes.

PR #388, opened 2024-03-10
downloader: local provider source with symlinks

VodBox:
To work with a local recording from OBS, requires that the output format is set to m3u8 + ts, and that the custom muxer settings includes ``hls_flags=+program_date_time``. ``hls_list_size=0`` will also ensure that OBS provides the full playlist and appends new segments without getting rid of old ones, which is useful for ensuring all segments are downloaded

### DIFF
--- a/downloader/downloader/main.py
+++ b/downloader/downloader/main.py
@@ -21,7 +21,7 @@ import common
 import common.dateutil
 import common.requests
 
-from .providers import URLProvider, TwitchProvider, YoutubeProvider
+from .providers import URLProvider, TwitchProvider, YoutubeProvider, LocalProvider
 
 
 segments_downloaded = prom.Counter(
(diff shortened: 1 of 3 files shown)


downloader/downloader/providers.py
```diff
@@ -195,3 +196,26 @@ def variant_name(variant):
...
+			playlist = hls_playlist.load(f.read().strip(), uri)
+		return playlist
+
+	def get_media_playlist_uris(self, qualities, session=None):
+		if qualities != ["source"]:
+			raise ValueError("Cannot provide non-source qualities")
+
+		files = os.listdir(self.directory)
+
+		for file in reversed(files):
```
ekimekim, line 217: What's the purpose of checking in specifically reversed order here? To get the "latest" version of the m3u8?

There's no guarentee that the order returned by `os.listdir` corresponds to anything in particular. In practice it will probably be the order files are created in, but it'd be better to explicitly sort by name:
```suggestion
		for file in sorted(files, reverse=True):
```

downloader/downloader/main.py
```diff
@@ -687,3 +699,69 @@ def stop():
 		worker.get() # re-raise error if failed
 
 	logging.info("Gracefully stopped")
+
+class SegmentLinker(SegmentGetter):
```
ekimekim, line 703: This is a style nitpick but I think the job of `SegmentGetter` and `SegmentLinker` is sufficiently different that `Linker` shouldn't inherit from `Getter`. Maybe it makes sense for them to have a common base class, or maybe everything HTTP-related should be moved out of `SegmentGetter` and then `SegmentGetter` is the common code and there's specific "local link" and "http fetch" providers (similar to how we have a twitch provider, youtube provider etc instead of a `TwitchStreamWorker` subclass of `StreamWorker`)

VodBox: I can do that. I mostly did it the way I did to reuse code with the smallest diff footprint, but I can definitely see that being better.

Issue #325, opened 2022-11-15
thrimbletrimmer: Disallow emoji in chapter titles


ekimekim:
As youtube seems to not accept them as chapters.
The exact allowed set of characters is not known.

PR #389, opened 2024-04-20
thrimbletrimmer: Validate chapter titles are ascii only

ekimekim:
We are not sure what characters are allowed in chapter titles.
Emoji seem to be disallowed. It is unknown whether things like accents or smart quotes are allowed.
To be conservative, we warn if there are any non-ascii characters in the chapter title.

Fixes https://github.com/dbvideostriketeam/wubloader/issues/325

### DIFF
--- a/thrimbletrimmer/scripts/edit.js
+++ b/thrimbletrimmer/scripts/edit.js
@@ -829,7 +829,7 @@ function validateChapterDescription(chapterDescField) {
 		// We don't know what chars are safe outside the ascii range, so we just warn on any of them.
 		// We know emoji are not safe.
 		chapterDescField.classList.add("input-error");
-		chapterDescField.title = "Chapter descriptions with non-ascii characters may cause issues, proceed with caution";
+		chapterDescField.title = "Chapter descriptions with non-ascii characters may cause issues; proceed with caution";
 	} else {
 		chapterDescField.classList.remove("input-error");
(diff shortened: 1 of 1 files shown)


ElementalAlchemist:
This'll work. My only complaint (because I *must* nitpick this, apparently) is that you used a comma splice in the error message; a semicolon is preferred in that position (or break it up in some other way you prefer).

thrimbletrimmer/scripts/edit.js
```diff
@@ -825,6 +825,11 @@ function validateChapterDescription(chapterDescField) {
 	if (chapterDesc.indexOf("<") !== -1 || chapterDesc.indexOf(">") !== -1) {
 		chapterDescField.classList.add("input-error");
 		chapterDescField.title = "Chapter description may not contain angle brackets (< or >)";
+	} else if (Array.from(chapterDesc).some(c => c.charCodeAt(0) > 127)) { // any char is non-ascii
+		// We don't know what chars are safe outside the ascii range, so we just warn on any of them.
+		// We know emoji are not safe.
+		chapterDescField.classList.add("input-error");
+		chapterDescField.title = "Chapter descriptions with non-ascii characters may cause issues, proceed with caution";
```
ekimekim, line 832:
```suggestion
		chapterDescField.title = "Chapter descriptions with non-ascii characters may cause issues; proceed with caution";
```

PR #390, opened 2024-05-30
Support video transitions

ekimekim:
This PR has quite a few changes to how cutting is done, in support of two new features:
- For full cuts, support multiple ranges (previously they only allowed one range)
- For fast, smart and full cuts, support cuts with transitions between ranges

This required rewriting ffmpeg handling considerably.
In the process I was able to build slightly higher level abstractions around calling ffmpeg which I think is a win overall.
The simple cut functions (eg. to build a waveform or extract a frame) are considerably easier.

It may be best to review this commit by commit - the first few commits are setting up the abstractions
I then use for rewriting the fast and full cuts.

We support all transitions built into the `xfade` filter in ffmpeg, including basics like `fade`
but also many kinds of wipe. We also have two custom wipes, a star wipe and a clock wipe.

Audio is cross-faded over the same time period as the video transition - we may revisit this in the future
but this was the simplest option.

The cutter needed almost no changes to pass in the transitions correctly. The restreamer learned a new option to the `cut` endpoint to also take a number of transitions.

### DIFF
--- a/common/common/segments.py
+++ b/common/common/segments.py
@@ -391,8 +391,13 @@ def ffmpeg_cut_segment(segment, cut_start=None, cut_end=None):
 	return subprocess.Popen(args, stdout=subprocess.PIPE)
 
 
+def ffmpeg_cut_one(segments, encode_args, output_file=subprocess.PIPE, input_args=[]):
+	"""Wrapper for ffmpeg_cut_many() with a simpler API for the single-input case."""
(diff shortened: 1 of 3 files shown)


ekimekim:
I'd rather not leave this sitting on top of master while other changes go in, so I'll just merge it now. I tested this when I first implemented it and it works. If we hit any problems it can be avoided by simply not doing any transitions.

Issue #376, opened 2024-01-14
Automated way to update schedulebot when shifts change


ekimekim:
Shifts can change for reasons beyond our control. We want to be able to handle:
- [x] Ordinary roles (editor, sheeter) being dropped/added/moved
- [ ] Backups/overrides for roles-by-shift (sups, chatops)

ekimekim:
Chatops is not a role-by-shift but is covered by normal schedule.
Sups and backups are grouped together in roles-by-shift so needing to hot update those isn't as important (ie. if the backup has taken over, they're already in the correct group anyway). And sups are already hot updated for the purposes of hourly reporting.
So we can punt on the last bits of roles-by-shift we'd want to auto-update, which is only (sups+backups) and wub wranglers.

PR #391, opened 2024-07-20
Support hot reload of Zulip Schedule

ZeldaZach:
- Move sheets API into common dir, since multi use
- Live download from Google Sheets using Config
- Falls back on old schedule if new one can't be downloaded for some reason

### DIFF
--- a/common/common/sheets.py
+++ b/common/common/sheets.py
@@ -1,7 +1,7 @@
 
 import logging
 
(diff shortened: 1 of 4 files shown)


ekimekim:
Closes https://github.com/dbvideostriketeam/wubloader/issues/376

ekimekim (approved):

Issue #372, opened 2023-11-18
manual links should optionally be modifyable


ekimekim:
if the manual link is to a video on our channel, we should be able to modify title/desc etc.
this is tricky because we'd need to start by detecting the *current* title etc.

PR #396, opened 2024-08-04
Allow the youtube-manual upload backend to modify videos

ekimekim:
So we can modify youtube videos we manually uploaded.
If the video isn't actually ours, it will error when the modify is attempted which is ok.


ekimekim:
I *think* this is all we need to make https://github.com/dbvideostriketeam/wubloader/issues/372 work, at least at a basic level. There might be some UI issues in thrimbletrimmer when trying to load a manually-uploaded video in order to modify it.

Issue #386, opened 2024-03-09
playlist manager - flag a video for "must be first/last"


ekimekim:
useful for supercuts

Issue #345, opened 2023-10-02
sheetsync: Streamlog support


ekimekim:

ekimekim:
Good progress was made but back-burnering for this year's run since we're not going to use streamlog.

PR #397, opened 2024-08-04
Allow setting a pinned first or last video in a playlist

ekimekim:
Co-authored by @ZeldaZach
Fixes https://github.com/dbvideostriketeam/wubloader/issues/386

This adds two configurable properties to a playlist, specifying a "first" and "last" event id.
Both are optional.

When present, they cause the video associated with that event to always be placed first or last in playlist order, overriding the normal "by start time" ordering and working retroactively (ie. they'll be moved to the front/back if not there already).
The primary use case for this is supercuts, which we generally want to put at the end of a playlist even if the associated event row coincides with the first instance of the thing being supercut.
It may also prove useful eg. if we get a clean version of a repeating intro, which would best go at the front of a playlist.

While conceptually simple, this change required some signifigant refactoring of `playlist_manager` to plumb more data to where it was needed, for example needing to cache the playlist entry ids instead of just video ids.
Along the way we do some refactoring for clarity, most notably splitting the use of ambiguous variables named `playlist` into either:
- `playlist` (A list of entries)
- `playlist_id` (A string identifier)
- `playlist_config` (A named tuple containing playlist tags and first/last event ids)

This change does not incorporate any way to *set* these first/last event ids, which would likely be in sheetsync - however I don't really want to make further changes there until https://github.com/dbvideostriketeam/wubloader/issues/345 is complete.

### DIFF
--- a/playlist_manager/playlist_manager/main.py
+++ b/playlist_manager/playlist_manager/main.py
@@ -142,27 +142,34 @@ class PlaylistManager(object):
 			logging.debug("All videos already correctly ordered in playlist, nothing to do")
 			return
 
-		# Refresh our playlist state, if necessary.
-		self.refresh_playlist(playlist_id)
+		for attempt in range(3):
+			# Refresh our playlist state, if necessary.
+			playlist = self.refresh_playlist(playlist_id)
 
-		# Get an updated list of new videos
-		playlist = self.get_playlist(playlist_id)
-		matching_video_ids = {video.video_id for video in matching}
-		playlist_video_ids = {entry.video_id for entry in self.get_playlist(playlist_id)}
-		# It shouldn't matter, but just for clarity let's sort them by event order
-		new_videos = sorted(matching_video_ids - playlist_video_ids, key=lambda v: v.start_time)
+			try:
+				# Perform any reorderings needed
+				reorderings = self.find_playlist_reorderings(videos, playlist, playlist_config)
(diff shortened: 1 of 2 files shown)


playlist_manager/playlist_manager/main.py
```diff
@@ -206,23 +326,48 @@ def insert_into_playlist(self, playlist, video_id, index):
...
 			)
-			if not resp.ok:
-				raise Exception("Failed to insert {video_id} at index {index} of {playlist} with {resp.status_code}: {resp.content}".format(
-					playlist=playlist, video_id=video_id, index=index, resp=resp,
-				))
+		if not resp.ok:
+			raise APIException("Failed to insert {video_id} at index {index} of {playlist} with {resp.status_code}: {resp.content}".format(
+				playlist=playlist_id, video_id=video_id, index=index, resp=resp,
+			), code=resp.status_code)
+		# TODO return entry_id from resp
```
ekimekim, line 333: will fix this before merging, just didn't have access to api docs on the plane

ZeldaZach:
Preliminary review, it seems fine. If it's functional, good to merge

ekimekim:
Still need to test. Unfortunately since it involves the youtube API it's a bit annoying to do.

chrusher:
Looks good to me.

ekimekim:
Will test properly as part of our all-up testing.

PR #401, opened 2024-08-24
Record in database when end time is "--"

ekimekim:
We need this so that reverse sync reproduces these values correctly.

To handle this in the database, we have a composite type (dashed: boolean, value: timestamp).
Value is always valid and is equivalent to the old timestamp column,
but must be equal to start_time if dashed is true.

The only place we directly reference this column outside sheetsync is thrimshim, where we
always consider the value only.

### DIFF
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -180,6 +180,8 @@ def get_row(ident):
 	response['id'] = str(response['id'])	
 	if response["video_channel"] is None:
 		response["video_channel"] = app.default_channel
+	# Unwrap end time (dashed, value) to just value
+	response["event_end"] = response["event_end"][1]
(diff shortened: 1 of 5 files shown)


thrimshim/thrimshim/main.py
```diff
@@ -408,6 +413,7 @@ def normalize(value):
 		return 'Sheet columns have changed since editing has begun. Please review changes\n' + changes, 409
 
 	if new_row['state'] == 'MODIFIED':
+		missing = []
```
ekimekim, line 416: Drive-by fix here for a bug I randomly spotted.
Previously we didn't define `missing` before using it here, but there happened to be a `missing` defined previously which would always be empty, so as long as no columns were ever missing it would still work.

chrusher:
Looks ok to me

Issue #405, opened 2024-08-30
Archive content from the image/links column


ekimekim:
If an image or video is referenced in the Image/Links column, we should save a copy as these tend to rot quickly. Related to https://github.com/dbvideostriketeam/wubloader/issues/403

PR #407, opened 2024-09-14
chat-archiver: Download media that is linked in chat

### LINKED ISSUE #403: chat-archiver: Save copies of images posted
Image links tend to rot quickly. To archive the chat we should also archive any images linked.

ekimekim:
Most of the logic here comes down to three thorny issues:
- How to identify links in chat
- Safely fetching from potentially-malicious URLs
- How to store that in the filesystem

We store things in such a way that it works with backfiller, so you can copy any data you miss
from other instances.

The purpose of all this is to provide an archive of images etc linked in chat *as they were at the time*, as links can rot quickly.
Special care is taken to re-check the content each time it's linked in case it returns a different result.

This may end up incurring a lot of disk usage. The expectation is that we'll only enable this for the main nodes.

Closes #403

### DIFF
--- a/common/common/media.py
+++ b/common/common/media.py
@@ -64,7 +64,7 @@ def download_media(
 	output_dir,
 	max_size=128*2**20, # 128MiB
 	timeout=60,
-	content_types=("image", "video"),
+	content_types=("image", "video", "application/pdf"),
 	max_redirects=5,
 	retries=3,
 	retry_interval=1,
--- a/common/common/media.py
+++ b/common/common/media.py
@@ -51,6 +51,9 @@ class BadScheme(Rejected):
 class WrongContent(Rejected):
 	"""Response was not a video or image"""
 
+class FailedResponse(Rejected):
+	"""Got a 4xx response, probably a bad link"""
+
 
 def check_for_media(output_dir, url):
 	"""Returns True if we have at least one version of content for the given url already."""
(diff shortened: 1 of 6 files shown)


ekimekim:
I ran this URL parser over last year's chat logs and it seems to work well. I compared it to a much looser version with a lot of false positives and it didn't have any false negatives (things it failed to find) while eliminating the vast majority of the false positives.

ekimekim:
Added a second part, also fixes #405

common/common/media.py
```diff
@@ -0,0 +1,247 @@
...
+	content_types=("image", "video", "application/pdf"),
+	max_redirects=5,
+	retries=3,
+	retry_interval=1,
+	chunk_size=64*1024, # 64KiB
+):
+	"""Make a GET request to a potentially malicious URL and download the content to file.
+	We check the following:
+	- That the host is a public IP
+	- That the response does not exceed given max size (default 100MB)
```
chrusher, line 79: The max size is set to 128 MB by default not 100 MB

docker-compose.jsonnet
```diff
@@ -21,7 +21,7 @@
     restreamer: true,
     backfiller: true,
     cutter: true,
-    sheetsync: false,
+    sheetsync: true,
```
chrusher, line 24: Should sheetsync be active by default?

chrusher (approved):
Looks good to me apart from a couple minor queries

common/common/media.py
ekimekim: ah good catch, that number changed a few times between writing the docs and the final thing.

docker-compose.jsonnet
ekimekim: whoops, this was testing code that leaked in

ekimekim:
I'm just going to merge this without fully testing it, a decision I could not possibly come to regret

PR #410, opened 2024-10-09
Add thrimbletrimmer UI for rich transitions

ekimekim:
It gets the list of available transitions from thrimshim,
and creates a drop-down + duration picker between each range.


ElementalAlchemist (approved):
I see you got fed up with my use of raw DOM APIs. (Not that I can blame you; they're quite verbose.)

Issue #377, opened 2024-01-14
thumbnails: Allow overriding template parameters in the editor


ekimekim:
Sometimes we require a "custom" thumbnail only because the default template settings don't quite capture the right part of the image.
We currently have a partial mitigation for this in the form of `foo-left` and `foo-right` templates, which capture slightly different screen areas and can be chosen between as needed.
A more general solution would be to allow the editor to override template settings on a per-video basis.
To be usable, this would likely require some sort of UI for seeing where on the screen is going to be grabbed. Or maybe just iterating on changing settings + hitting show preview is good enough?
On the backend this should be an easy enough change, though it will require a new database column to possibly contain template settings.

Issue #366, opened 2023-11-15
thumbnails: Allow combining frame with one-off uploaded template


ekimekim:
It is often the case that a custom one-off thumbnail is still made up of a "border" around the outside + a still frame from the video. In these cases it is annoying extra work to download a frame and essentially manually perform the templating process. But we also don't want to add the one-off as a template as this clutters up the templates list.

We should be able to accept a custom template image in thrimbletrimmer, which is then applied to the chosen frame to make a thumbnail.
This could be done without signifigant backend changes by implementing it in restreamer, having thrimbletrimmer download the image, then present it as a fully custom thumbnail when uploading. Or we could add a new `thumbnail_mode`, though this would mean a) needing to stash the templating parameters somewhere (in the template "name"?), b) putting the custom template in the thumbnail image column is a problem since this column *should* contain the generated image. We would probably instead want a new column.

chrusher:
I think generating the template in the restreamer from thrimbletrimmer then treating it as a custom template is the better way.

ekimekim:
Agreed. Are you taking this as part of the thumbnail work?

chrusher:
Yes. I might as well do all the thumbnail work as one project.

Issue #358, opened 2023-11-12
Automate deployments of thumbnail templates 


chrusher:
It would be nice if the thumbnail templates could be updated on one node and the other edit nodes could automatically update. Existing `backfiller` code could be adapted but special handling will be need for when a template is updated rather than added.

ekimekim:
Bonus work: Making it possible for non-technical folks to upload/update templates.

ekimekim:
Thinking about it, it may be easier to store the templates in the database. They are not large, nor accessed frequently. In addition, thrimshim needs to know which ones exist and thrimshim is designed to not nessecarily have disk access. This solves the distribution problem (even for volunteer edit nodes), though we would need to make some tooling for template management. But we want to do that anyway, so non-technical folks can manage it.

chrusher:
Yeah that seems like a reasonable solution.

For template management, we would want to add, replace, rename and delete templates as well as setting and modifying the template parameters and any template attribution (see issue #380). I assume a web form to do this should not be too difficult. We likely want to put authentication on the web form so only the art team and the wub wranglers can modify things.

ekimekim:
An additional thought on this: Right now the default thumbnail template is selected via a hard-coded list inside thrimshim. This has already broken things once because we renamed the templates and didn't update thrimshim.
A better solution would probably be to make "default template for each shift" selectable in whatever UI we end up with.

This is a bonus though, our current solution does work.

chrusher:
Simplest way to do that I would think is to have a `default_templates` table in the database that the wubloader wranglers could edit as needed. We could also have a `is_omega` flag we can flip when it reaches that part of the run.

PR #413, opened 2024-10-20
Thumbnail Template Management

chrusher:
Rather than store thumbnail templates as files that have to manually copied to the edit nodes, store the thumbnails and associated metadata in the database (#358). Additionally, the crop and location thumbnail parameters can be overridden (#377) and a one-off template can be combined with a video frame to produce a custom thumbnail (#366). 

This is a work in progress and has not been tested (although it does build).

### DIFF
--- a/common/common/images.py
+++ b/common/common/images.py
@@ -1,32 +1,20 @@
 
-import json
-import os
 import sys
 from io import BytesIO
 
 from PIL import Image
 
+
 """
-A template is two files:
-	NAME.png
-	NAME.json
-The image is the template image itself.
-The JSON file contains the following:
-	{
-		crop: BOX,
-		location: BOX,
-	}
-where BOX is a 4-tuple [left x, top y, right x, bottom y] describing a rectangle in image coordinates.
+A thumbnail template consists of a byte stream representation of a PNG template image along with two 4-tuples [left x, top y, right x, bottom y] describing the crop and location of the frame. 
 
 To create a thumbnail, the input frame is first cropped to the bounds of the "crop" box,
 then resized to the size of the "location" box, then pasted underneath the template image at that
 location within the template image.
 
-For example, a JSON file of:
-	{
-		"crop": [50, 100, 1870, 980],
-		"location": [320, 180, 1600, 900]
-	}
+For example
+crop = (50, 100, 1870, 980)
+location = (320, 180, 1600, 900)
 would crop the input frame from (50, 100) to (1870, 980), resize it to 720x1280,
 and place it at (320, 180).
 
@@ -34,18 +22,13 @@ If the original frame and the template differ in size, the frame is first resize
 This allows you to work with a consistent coordinate system regardless of the input frame size.
 """
 
-def compose_thumbnail_template(base_dir, template_name, frame_data):
-	template_path = os.path.join(base_dir, "thumbnail_templates", f"{template_name}.png")
-	info_path = os.path.join(base_dir, "thumbnail_templates", f"{template_name}.json")
+def compose_thumbnail_template(template_data, frame_data, crop, location):
 
-	template = Image.open(template_path)
 	# PIL can't load an image from a byte string directly, we have to pretend to be a file
+	template = Image.open(BytesIO(template_data))
 	frame = Image.open(BytesIO(frame_data))
 
-	with open(info_path) as f:
-		info = json.load(f)
-	crop = info['crop']
-	loc_left, loc_top, loc_right, loc_bottom = info['location']
+	loc_left, loc_top, loc_right, loc_bottom = location
 	location = loc_left, loc_top
 	location_size = loc_right - loc_left, loc_bottom - loc_top
 
@@ -72,14 +55,15 @@ def compose_thumbnail_template(base_dir, template_name, frame_data):
 	buf.seek(0)
 	return buf.read()
 
-
-def cli(template_name, image, base_dir="."):
-	with open(image, "rb") as f:
-		image = f.read()
-	thumbnail = compose_thumbnail_template(base_dir, template_name, image)
+def cli(template, frame, crop, location):
+	with open(template, "rb") as f:
+		template = f.read()
+	with open(frame, "rb") as f:
+		frame = f.read()
+	thumbnail = compose_thumbnail_template(template, frame, crop, location)
 	sys.stdout.buffer.write(thumbnail)
 
 
 if __name__ == '__main__':
 	import argh
-	argh.dispatch_command(cli)
+	argh.dispatch_command(cli)		
--- a/common/common/images.py
+++ b/common/common/images.py
@@ -4,25 +4,43 @@ from io import BytesIO
 
 from PIL import Image
 
+from common import database
 
-"""
-A thumbnail template consists of a byte stream representation of a PNG template image along with two 4-tuples [left x, top y, right x, bottom y] describing the crop and location of the frame. 
 
-To create a thumbnail, the input frame is first cropped to the bounds of the "crop" box,
-then resized to the size of the "location" box, then pasted underneath the template image at that
-location within the template image.
+def get_template(dbmanager, name, crop=None, location=None):
+	"""Fetch the thumbnail template and any missing parameters from the database"""
+	with dbmanager.get_conn() as conn:
+		query = """
+			SELECT image, crop, location FROM templates WHERE name = %s
+		"""
+		results = database.query(conn, query, name)
+		row = results.fetchone()
+		if row is None:
+			raise ValueError('Template {} not found'.format(name))
+		row = row._asdict()
+		if not crop:
+			crop = row['crop']
+		if not location:
+			location = row['location']
 
-For example
-crop = (50, 100, 1870, 980)
-location = (320, 180, 1600, 900)
-would crop the input frame from (50, 100) to (1870, 980), resize it to 720x1280,
-and place it at (320, 180).
+		return row['image'], crop, location
 
-If the original frame and the template differ in size, the frame is first resized to the template.
-This allows you to work with a consistent coordinate system regardless of the input frame size.
-"""
 
 def compose_thumbnail_template(template_data, frame_data, crop, location):
+	"""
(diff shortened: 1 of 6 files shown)


thrimshim/thrimshim/main.py
```diff
@@ -46,13 +46,50 @@ def _start_response(status, headers, exc_info=None):
...
 
+def authenticate_artist(f):
+	""""Authenticate a token against the database to authenticate an artist
 
-def authenticate(f):
-	"""Authenticate a token against the database.
+	Reference: https://developers.google.com/identity/sign-in/web/backend-auth"""
+    
+	@wraps(f)
+	def artist_auth_wrapper(*args, **kwargs):
```
ekimekim, line 55: Rather than repeating this entire function with `artist` instead of `editor`, could we have one version of the function and pass in the role that we want to test?

postgres/schema.sql
```diff
@@ -89,6 +89,8 @@ CREATE TABLE events (
 		OR thumbnail_mode = 'NONE'
 		OR thumbnail_last_written IS NOT NULL
 	),
+	thumbnail_crop INTEGER[], -- left, upper, right, and lower pixel coordinates to crop the selected frame
+	thumbnail_location INTEGER[], -- left, top, right, bottom pixel coordinates to position the cropped frame
```
ekimekim, line 93: It would be good to verify that these values are only ever 4 items, if they're set at all. eg. `CHECK cardinality(thumbnail_crop) = 4 OR thumbnail_crop IS NULL`

postgres/schema.sql
```diff
@@ -173,3 +175,16 @@ CREATE TABLE bus_data (
...
+-- attribution: any attribution to be auto included in the video description. If empty, do not add an attribution
+-- crop: left, upper, right, and lower pixel coordinates to crop the selected frame
+-- location: left, top, right, bottom pixel coordinates to position the cropped frame
+CREATE TABLE templates (
+	name TEXT PRIMARY KEY,
+	image BYTEA NOT NULL,
+	description TEXT NOT NULL DEFAULT '',
+	attribution TEXT NOT NULL DEFAULT '',
+	crop INTEGER[] NOT NULL,
+	location INTEGER[] NOT NULL
```
ekimekim, line 189: as above, we should check the lengths on these: `CHECK cardinality(crop) = 4`

thrimshim/thrimshim/main.py
```diff
@@ -607,6 +608,171 @@ def _write_audit_log(conn, ident, api_action, editor, old_row=None, new_row=None
...
+
+	with app.db_manager.get_conn() as conn:
+
+		#check if name is already in the database
+		query = sql.SQL("""
+			SELECT name FROM events WHERE name = %s 
+		""")
+		results = database.query(conn, query, new_template['name'])
+		if results.fetchone() is not None:
+			return 'Template with name {} already exists'.format(new_template['name']), 400
```
ekimekim, line 688: Nitpicking: we could use `409` (Conflict) instead of `400` here.

thrimshim/thrimshim/main.py
```diff
@@ -607,6 +608,171 @@ def _write_audit_log(conn, ident, api_action, editor, old_row=None, new_row=None
...
+		query = sql.SQL("""
+			INSERT INTO templates ({})
+			VALUES ({})
+		""").format(
+				sql.SQL(", ").join(sql.Identifier(column) for column in columns),
+				sql.SQL(", ").join(database.get_column_placeholder(column) for column in columns),
+			)
+		database.query(conn, query, **new_template)
+	
+	return ''
```
ekimekim, line 699: *very* nitpicking: We could return a `201` Created here instead of a `200` OK.

thrimshim/thrimshim/main.py
```diff
@@ -607,6 +608,171 @@ def _write_audit_log(conn, ident, api_action, editor, old_row=None, new_row=None
...
+
+@app.route('/thrimshim/update-template/<name>', methods=['POST'])
+@request_stats
+def update_template(name, artist=None):
```
ekimekim, line 703: This function is 90% the same as the create function, could we split out and reuse the common parts for input parsing, etc?

thrimshim/thrimshim/main.py
```diff
@@ -607,6 +608,171 @@ def _write_audit_log(conn, ident, api_action, editor, old_row=None, new_row=None
...
+		results = database.query(conn, query, ident)
+		row = results.fetchone()
+		if row is None:
+			return 'Event {} not found'.format(ident), 404
+		event = row._asdict()
+
+		if event['thumbnail_mode'] != 'NONE' and event['thumbnail_image']:
+			return flask.Response(event['thumbnail_image'], mimetype='image/png')
+		else:
+			return '', 200
```
ekimekim, line 772: I don't think returning a "success" with no data is the right thing here semantically. Maybe a 404?

common/common/images.py
```diff
@@ -72,14 +55,15 @@ def compose_thumbnail_template(base_dir, template_name, frame_data):
...
+	with open(frame, "rb") as f:
+		frame = f.read()
+	thumbnail = compose_thumbnail_template(template, frame, crop, location)
 	sys.stdout.buffer.write(thumbnail)
 
 
 if __name__ == '__main__':
 	import argh
-	argh.dispatch_command(cli)
+	argh.dispatch_command(cli)
```
ekimekim, line 69: trailing whitespace

restreamer/restreamer/main.py
```diff
@@ -493,23 +480,43 @@ def get_frame(channel, quality):
...
-@has_path_args
 def get_thumbnail(channel, quality):
 	"""
 	Returns a PNG image which is a preview of how a thumbnail will be generated.
-	Params:
-		timestamp: Required. The frame to use as the thumbnail image.
-			Must be in ISO 8601 format (ie. yyyy-mm-ddTHH:MM:SS) and UTC.
-		template: Required. The template name to use.
-			Must be one of the template names (without file extension) as returned
-			by GET /files/thumbnail_templates
```
ekimekim, line 505: Replace with updated documentation, please.

restreamer/restreamer/main.py
```diff
@@ -493,23 +480,43 @@ def get_frame(channel, quality):
 
 @app.route('/thumbnail/<channel>/<quality>.png')
 @request_stats
-@has_path_args
```
ekimekim, line 496: This function still has path args (`channel` and `quality`) which need to be checked by `has_path_args` for security.

restreamer/restreamer/main.py
```diff
@@ -493,23 +480,43 @@ def get_frame(channel, quality):
...
-			Must be one of the template names (without file extension) as returned
-			by GET /files/thumbnail_templates
 	"""
-	template_name = request.args['template']
-	template_path = os.path.join(app.static_folder, "thumbnail_templates", f"{template_name}.png")
-	if not os.path.exists(template_path):
-		return "No such template", 404
 
-	timestamp = dateutil.parse_utc_only(request.args['timestamp'])
+	template_params = request.json
```
ekimekim, line 488: This is a GET request, `request.json` shouldn't be used as it reads json from the request body, and while GET requests with a body may technically work most of the time, it's officially disallowed by the standard and not everything will support it.

I think the cleanest option here would be to have two versions of this function:
- A `GET` endpoint which uses query params to take a timestamp, template name and crop metadata
- A `POST` endpoint which uses query params to take a timestamp and crop metadata, but gets the template image from the request body

This also saves having to base64 the image template when sending it, which makes it a bit nicer to use from `curl` etc.

So something like:
```
@app.route("/thumbnail/<channel>/<quality>.png")
@request_stats
@has_path_args
def get_thumbnail_named_template(channel, quality):
    template = get_template_from_db(request.args["template"])
    return get_thumbnail(channel, quality, template)

@app.route("/thumbnail/<channel>/<quality>.png")
@request_stats
@has_path_args
def get_thumbnail_uploaded_template(channel, quality):
    template = request.body
    return get_thumbnail(channel, quality, template)
```

ekimekim (commented):
Looks good so far. A few nitpicks, and some larger notes around the api of `get_thumbnail`.

thrimshim/thrimshim/main.py
chrusher: This is me not understanding decorators very well and being too lazy to figure out how to write a decorator that you can pass an argument to.

restreamer/restreamer/main.py
chrusher: I assume the second one should start `@app.route("/thumbnail/<channel>/<quality>.png" , methods=['POST'])` otherwise they have the same route.

ekimekim: yes, whoops.

thrimshim/thrimshim/main.py
ekimekim: You could still have two seperate decorators that just call out to a common function.
Or if you want a decorator that takes an argument, here is the standard idiom:
```
def mydecorator(arg):
    def mydecorator_generated(f):
        @wraps(f)
        def mydecorator_wrapper(*args, **kwargs):
            ...
        return mydecorator_wrapper
    return mydecorator_generated
```
It's a bit convoluted but the idea is that when you call `mydecorator(arg)`, it returns the actual decorator which has the arg "baked in", and the decorator then returns the wrapped function.

You'd then use it like:
```
@mydecorator(arg)
def wrapped_func():
    ...
```

thrimshim/thrimshim/main.py
chrusher: I am splitting the common input parsing off into its own function but keeping the database interactions in the individual functions.

thrimshim/thrimshim/main.py
chrusher: I am going with two separate decorators that share a single function.

ekimekim (approved):
New changes look good

ekimekim:
I've made the minimal change to thrimbletrimmer to keep it working using the new thumbnail apis, and am going to merge this so that I can get the database changes into master.

Just to check my understanding, after this PR is merged, we have the following left to do:
- [ ] Auto-attribution needs to actually happen (added to description in thrimshim)
- [ ] UI for template management
- [ ] UI for one-off templating
- [ ] Better way of doing default templates

Did I miss anything?

PR #415, opened 2024-10-22
Document remaining transitions, deprecate fadefast and fadeslow.

Tantusar:
It turns out that `fadefast` and `fadeslow` both take about twice as long as `fade` to do a job so similar there's no good reason to keep either in our accepted transitions list, especially when the former is so misleadingly named. (Amusingly, in my testing, `fadefast` was actually the slower of the two.)


Tantusar:
I have no idea what `distance` is doing either, by the way. Was making videos of the various transitions for an appendix of the manual, and it helped me work out everything _except_ `distance`.

https://github.com/user-attachments/assets/3e15c858-e967-4608-8925-be7f3af4843b

ekimekim (approved):
Thanks for this.
fadefast/fadeslow is especially bizzare because the fade duration is meant to be a seperate parameter...
I *think* distance is something to do with colors? :shrug: We can just ignore it.

PR #417, opened 2024-10-23
thrimbletrimmer: Add support for advanced crop options

ekimekim:
- Advanced options are in an expandable section
- On load + change of template, set from template defaults
- On load, set from video if not null
- Show only in template mode
- Use when previewing image
- Send when submitting

### DIFF
--- a/thrimbletrimmer/edit.html
+++ b/thrimbletrimmer/edit.html
@@ -252,6 +252,31 @@
 							id="video-info-thumbnail-time-play"
 						/>
 					</div>
+					<div class="video-info-thumbnail-mode-options" id="video-info-thumbnail-position-options">
+						<details>
+							<summary>Advanced Templating Options</summary>
+							Crop specifies the region of the video frame to capture. <br/>
+							Location specifies the region within the template image where the cropped image will be placed. <br/>
+							Regions are given as pixel coordinates of the top-left and bottom-right corners. <br/>
+							Note that if the regions are different sizes, the image will be stretched. <br/>
+
+							Crop:
+							<input type="text" class="video-info-thumbnail-position" id="video-info-thumbnail-crop-0" />
+							<input type="text" class="video-info-thumbnail-position" id="video-info-thumbnail-crop-1" />
+							to
+							<input type="text" class="video-info-thumbnail-position" id="video-info-thumbnail-crop-2" />
+							<input type="text" class="video-info-thumbnail-position" id="video-info-thumbnail-crop-3" />
+							<br/>
+
+							Location:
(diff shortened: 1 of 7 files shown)


chrusher (approved):
Looks good to me.

thrimbletrimmer/edit.html
```diff
@@ -252,6 +252,31 @@
...
+							<br/>
+
+							Location:
+							<input type="text" class="video-info-thumbnail-position" id="video-info-thumbnail-location-0" />
+							<input type="text" class="video-info-thumbnail-position" id="video-info-thumbnail-location-1" />
+							to
+							<input type="text" class="video-info-thumbnail-position" id="video-info-thumbnail-location-2" />
+							<input type="text" class="video-info-thumbnail-position" id="video-info-thumbnail-location-3" />
+							<br/>
+
```
ElementalAlchemist, line 278: Did you intentionally miss the closing `</details>` tag?

thrimbletrimmer/scripts/edit.js
```diff
@@ -267,27 +271,50 @@ window.addEventListener("DOMContentLoaded", async (event) => {
...
 			const imageTemplate = document.getElementById("video-info-thumbnail-template").value;
-			imageElement.src = `/thumbnail/${globalStreamName}/source.png?timestamp=${imageTime}&template=${imageTemplate}`;
+			const [crop, loc] = getTemplatePosition();
+			const queryParts = [
+				`timestamp=${imageTime}`,
+				`template=${imageTemplate}`,
+				`crop=${crop.join(",")}`,
+				`location=${loc.join(",")}`,
+			];
+			imageElement.src = `/thumbnail/${globalStreamName}/source.png?${queryParts.join("&")}`;
```
ElementalAlchemist, line 281: We should make sure we document somewhere that template names may not contain ampersands, unless we intend on URI-escaping them.

ElementalAlchemist (changes requested):

thrimbletrimmer/scripts/edit.js
chrusher: We can enforce template names being limited to `[A-Za-z0-9_-]` when creating or modifying them by adding a check to `thrimshim.validate_template`.

thrimbletrimmer/edit.html
ekimekim: fixed

thrimbletrimmer/scripts/edit.js
ekimekim: I've replaced all ad-hoc query string building with proper usage of `URLSearchParams()` which will automatically do correct encoding.

PR #418, opened 2024-10-24
thrimletrimmer: Support one-off overlays for thumbnails

ekimekim:
By picking the "one-off overlay" option for a thumbnail, you swap
specifying a template name for being able to upload a one-off template that
is then combined with the requested frame.

The rendering is done by restreamer, and we do it explicitly whenever a)
Generate Thumbnail Preview is pressed b) The video is submitted

The rendered thumbnail is then included in the submission as a "custom"
thumbnail.

The default thumbnail template params (crop and location) do not change 
when this mode is selected, so they'll effectively be the default params of
the previously-selected template. In most cases this will be what you want
since almost all our templates share the same params, and custom one-offs
will too.


chrusher (approved):
Looks ok to me

PR #425, opened 2024-10-29
Thrimbletrimmer: Add JS cropping tool to advanced template options

### LINKED ISSUE #424: Implement a GUI for custom thumbnail cropping
When Sokar was showing us the editor updates, I suggested a GUI to help set the advanced thumbnail cropping options. (The new fields to set the x/y coordinates for the section of a frame to be used, and the section of the thumbnail for it to be scaled to.) This would help because it is difficult to estimate pixel coordinates without help from an image editor of some sort, and I have seen JS libraries that provide some image cropping abilities in the past.

dcollinsn:
This patch adds a JS tool called Jcrop which allows click-and-drag selection of parts of an image, so that editors using the custom thumbnail features can more easily select the parts of the video frame they want to use, and the location of the template where they want to place it. The editor can click a button to preview the two source images (video frame and template), which also creates a Jcrop stage for each. The canvas is pre-populated with the coordinates from the text input fields. Whenever the text fields, or the Jcrop selection, is changed by the editor, the other is updated automatically. 

Resolves #424


dcollinsn:
This is mostly a tech demo, I'm not sure if you'd prefer a different approach or are aware of a better way to do this. There are a few potential TODOs:

- Tagging the text fields as number inputs rather than text might allow browsers to provide handy up/down arrows and hotkey support for easier fine-tuning.
- A "lock aspect ratio" checkbox to avoid stretching would be handy. The template's Jcrop stage can be locked to match the aspect ratio of the selected area of the video frame.
- I didn't test this with any mode other than "use video frame in image template".
- Jcrop was the best choice I could easily find, but I already had to work around two cases where the documentation doesn't match the behavior (you can't delete the last widget from a stage, and the "crop.move" event does nothing), so it may have a high technical debt if it is unstable / not accurately documented.
- The "active" crop widget (at least in Firefox) has an extra border around it which makes precise selection difficult. I can't find the CSS that is doing this, but we may want to find a way to override it.

chrusher:
Currently we resize the frame from 1920 x 1080 to 1280 x 720 to match the template before doing the crop and location shift. This is under the assumption that the crop and location shift do not introduce any more scaling. This is fine for the default parameters but if we are punching in on something this results in double scaling.

dcollinsn:
That's true, but I don't think it's unique to this patch. The double-scaling was already possible with just the numerical inputs. However, this UI does make it easier to switch to cropping the full 1080p frame now, you don't have to do the mental gymnastics of guessing pixel offsets with a 1080 frame in a 720 template.

This particular library does use/return pixel values based on the displayed size of the image, so we just need to know how big the image is being displayed, so we can divide/multiply by the right amount (2 for 720, 3 for 1080) to get pixel coordinates in the source image.

thrimbletrimmer/edit.html
```diff
@@ -18,6 +18,8 @@
 			content="345276493482-r84m2giavk10glnmqna0lbq8e1hdaus0.apps.googleusercontent.com"
 		/>
 		<script src="https://apis.google.com/js/platform.js?onload=onGLoad" async defer></script>
+		<link rel="stylesheet" href="https://unpkg.com/jcrop/dist/jcrop.css">
+		<script src="https://unpkg.com/jcrop"></script>
```
ElementalAlchemist, line 22: Can we ship the script ourselves (like we do with Luxon and hls.js) instead of relying on this site?

thrimbletrimmer/scripts/edit.js
```diff
@@ -15,6 +15,10 @@ const PAGE_STATE = {
 	CONFIRMING: 3,
 };
 
+// References to Jcrop "stages" for the advanced thumbnail editor crop tool
+var videoFrameStage;
+var templateStage;
```
ekimekim, line 20: nitpick: Prefer `let` for new variable declarations, unless we need the specific weirdness of `var` for some reason.
```suggestion
let videoFrameStage;
let templateStage;
```

thrimbletrimmer/scripts/edit.js
```diff
@@ -2156,3 +2196,70 @@ function canEditMetadata() {
...
+	const templateFieldX2 = document.getElementById("video-info-thumbnail-location-2");
+	const templateFieldY2 = document.getElementById("video-info-thumbnail-location-3");
+	const templateRect = Jcrop.Rect.create(templateFieldX1.value/2, templateFieldY1.value/2, (templateFieldX2.value-templateFieldX1.value)/2, (templateFieldY2.value-templateFieldY1.value)/2);
+	if (templateStage.active == null) {
+		templateStage.newWidget(templateRect);
+	} else {
+		templateStage.active.pos = templateRect;
+		templateStage.active.render();
+	}
+}
```
ekimekim, line 2265: Please ensure your editor adds trailing newlines to files.
```suggestion
}
```

ekimekim (approved):
Apart from vendoring the dependency, this looks good to go. And is way less code than I expected.

ekimekim (approved):

ekimekim:
I will also update the templating code to not do the resize before cropping.

PR #426, opened 2024-10-30
playlist-audit: Script to help with playlist auditing

ekimekim:
As per Sokar's request.
It's not entirely ideal that we're copying the "should be in playlists" criteria here
and in the playlist manager, but it's simple enough (is DONE, is public, matching upload location, tags are a superset of playlist tags) that it should be fine.


chrusher (approved):
Looks ok to me

PR #427, opened 2024-10-31
download_media: Special casing for imgur

ekimekim:
Handle imgur links, including links to albums/galleries by downloading the source image
(or for albums, all images) instead of the directly linked url.

### DIFF
--- a/common/common/media.py
+++ b/common/common/media.py
@@ -20,6 +20,11 @@ from . import atomic_write, ensure_directory, jitter, listdir
 from .stats import timed
 
 
+# Lots of things will tell you to go away if you don't look like a browser
+# (eg. imgur)
+USER_AGENT = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/130.0.0.0 Safari/537.36"
+
+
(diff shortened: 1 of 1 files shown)


ekimekim:
This code is not yet tested.

common/common/media.py
```diff
@@ -245,3 +273,110 @@ def _request(url, max_size, content_types):
...
+			These map to actual media and are stored in the usual way.
+		imgur.com/a/ID
+		imgur.com/gallery/ID
+			These map to collections of media.
+			Under the original URL we store a json file that lists imgur.com/ID urls
+			of the contents of the collection. Those urls are then downloaded and stored
+			in the usual way.
+	"""
+	parsed = urllib.parse.urlparse(url)
+	if parsed.hostname not in ("imgur.com", "i.stack.imgur.com"):
```
dcollinsn, line 293: What about `i.imgur.com`? (Mentioned in the docstring but not here)

ekimekim: `i.imgur.com` requires no special handling (it's a direct link to the image) so we don't need to handle it here.

ekimekim: I can clarify that in the docstring, you're right it does say this function "handles" that case.

chrusher (approved):
Looks ok to me

PR #433, opened 2024-11-01
Attach validation to the first chapter marker description field

### LINKED ISSUE #429: thrimbletrimmer: First chapter marker doesn't turn red with illegal characters
All the other ones already do.

dcollinsn:
All of the other chapter markers get this event listener when they are created, but since this chapter marker already exists in the raw HTML, the event listener needs to be created when the page loads.

Resolves #429

### DIFF
--- a/thrimbletrimmer/scripts/edit.js
+++ b/thrimbletrimmer/scripts/edit.js
@@ -222,6 +222,9 @@ window.addEventListener("DOMContentLoaded", async (event) => {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


ElementalAlchemist (approved):
lol, whoops

It's amazing that no one noticed this for 3 years

ekimekim:
Failing builds is a known issue whne someone without write access to the repo makes a PR. Still haven't had time to handle it properly.

PR #434, opened 2024-11-01
Database fixes

chrusher:
Fixing a typo in the schema and updating from `editors` to `roles` in the setup script


ekimekim (approved):
I didn't even realise we had a thing for loading in roles...I've always done it manually via psql once the DB is up.

PR #436, opened 2024-11-01
Fixed missing authentication decorators

chrusher:
Added missing authentication decorators on `thrimshim` calls that modify the `templates` database table

Thanks to @ElementalAlchemist for noticing this


ekimekim (approved):
oof, good catch.

PR #439, opened 2024-11-02
Special-case "latest" in CACHE_FROM in build script

HubbeKing:
Try to fix build caching by just pulling latest in builds and caching from latest


HubbeKing:
Builds should now pull "latest" before building and use the layers of that for caching, which _should_ speed up builds???

PR #440, opened 2024-11-03
Allow thumbnail templates to be updated without providing an image

chrusher:
To cut down on overhead including an new image is now optional when updating a template.

### DIFF
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -644,9 +644,12 @@ def get_template_metadata(name):
 		logging.info('Thumbnail metadata of {} fetched'.format(name))
 		return json.dumps(row._asdict())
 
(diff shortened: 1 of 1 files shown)


ekimekim (approved):

thrimshim/thrimshim/main.py
```diff
@@ -656,15 +659,16 @@ def validate_template(new_template):
 	for extra in extras:
 		del new_template[extra]
```
ElementalAlchemist, line 660: Wouldn't this mean we can *never* update the image due to it being removed as a "non-required" field?

chrusher: If `image` is a field in the `new_template` it gets added to the list of columns, even if `require_image` is `False`.

PR #441, opened 2024-11-03
Get info on the thumbnails of all videos with a given tag

chrusher:


thrimshim/thrimshim/main.py
```diff
@@ -763,6 +763,26 @@ def get_thumbnail(ident):
...
+	"""
+	with app.db_manager.get_conn() as conn:
+		query = """
+			SELECT id, video_id, upload_location, thumbnail_mode, thumbnail_template
+			FROM events
+			WHERE %s = ANY (tags) AND state = 'DONE' AND public
+		"""
+		results = database.query(conn, query, tag)
+		videos = [row._asdict() for row in results]
+		if not videos:
```
ekimekim, line 781: I think it's more correct to say that a tag with no matching videos returns an empty list of videos. No results, not "missing results".

ekimekim (approved):
Functionally looks good. Where are we expecting to use this? And would it be better to take a playlist id and return all videos that match that playlist?

chrusher:
I was originally going to use playlist ids but I switched to the tags thinking they would be more human readable (and sidestepping dealing with a playlist having multiple tags) but I forgot that tags often contain characters that make URLs messy (spaces and sometimes emoji).

ekimekim:
As per discussion in zulip it seems that Sokar doesn't need this (everything he needs is already covered by ./playlist-audit) so I think we can just put aside this code for now. We can always bring it back if a use-case comes up.

PR #443, opened 2024-11-03
trying to fix caching stuff

ekimekim:
Refactor build script to actually have a working cache using buildah


HubbeKing (approved):
Looks good! Should we also add a check to the main.yml workflow, so that it sets CACHE=readonly if it's a pull request? That way it should only populate the cache when we're doing known-good stuff from actual repo contributors with write access.

PR #447, opened 2024-11-04
Tag Based Default Thumbnail Templates

chrusher:
Modifies `thrimshim.get_row` to choose the default thumbnail template based on tags and modifies `sheetsync` to determine the shift of an event and added it as a tag.

### DIFF
--- a/common/common/shifts.py
+++ b/common/common/shifts.py
@@ -1,5 +1,7 @@
 
 import datetime
+import urllib
+import zoneinfo
 
 from common import dateutil
 from common.requests import InstrumentedSession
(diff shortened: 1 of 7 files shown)


thrimshim/thrimshim/main.py
```diff
@@ -236,28 +236,30 @@ def get_row(ident):
...
+			conn = app.db_manager.get_conn()
+			query = """
+				SELECT name, default_template
+				FROM playlists
+					"""
+			results = database.query(conn, query)
+			default_templates = {}
+			for row in results:
+				row = row._asdict:
+				if row['name'] and row['default_template']:
```
ekimekim, line 255: instead of filtering things out here, you can add a WHERE clause to the sql, eg. `WHERE default_template IS NOT NULL`.

thrimshim/thrimshim/main.py
```diff
@@ -236,28 +236,30 @@ def get_row(ident):
...
+					default_templates[row['name']] = row['default_template']
```
ekimekim, line 256: IIUC you want the first tag of the playlist, not the playlist name? names will not always be the same as tags.

thrimshim/thrimshim/main.py
```diff
@@ -236,28 +236,30 @@ def get_row(ident):
...
+		if response['thumbnail_template'] is None
+			conn = app.db_manager.get_conn()
+			query = """
+				SELECT name, default_template
+				FROM playlists
+					"""
+			results = database.query(conn, query)
+			default_templates = {}
+			for row in results:
+				row = row._asdict:
```
ekimekim, line 254: Generally it's easier to work with the row as a namedtuple instead of a dict, unless you need to pass it to something that expects a dict, or iterate over keys etc.
So for example, without this line you could write `row.name` instead of `row['name']`.

thrimshim/thrimshim/main.py
```diff
@@ -236,28 +236,30 @@ def get_row(ident):
...
+				if row['name'] and row['default_template']:
+					default_templates[row['name']] = row['default_template']
+
+			for tag in response['tags'][::-1]:
```
ekimekim, line 258: We should explain why we prioritize later tags (because implicit tags are put at the start)

thrimshim/thrimshim/main.py
```diff
@@ -236,28 +236,30 @@ def get_row(ident):
...
+				if tag in default_templates:
+					response['thumbnail_template'] = default_templates[tag]
```
ekimekim, line 260: Right now the loop will continue after finding a matching tag, and overwrite the result with a new result.
You probably want a `break` here.

thrimshim/thrimshim/main.py
```diff
@@ -236,28 +236,30 @@ def get_row(ident):
...
-				pst_hour = (start.hour - 8) % 24
-				shift = int(pst_hour / 6)
-				response['thumbnail_template'] = DEFAULT_TEMPLATES[shift]
+
+		# use tags to determine default thumbnail template
+		if response['thumbnail_template'] is None
+			conn = app.db_manager.get_conn()
+			query = """
+				SELECT name, default_template
+				FROM playlists
```
ekimekim, line 249: Also mentioned below, we want tag instead of name and to exclude nulls here.
We should also exclude any playlist without exactly one tag.
Putting it all together:
```suggestion
				SELECT tags[1] as tag, default_template
				FROM playlists
				WHERE cardinality(tags) = 1 AND default_template IS NOT NULL
```

sheetsync/setup.py
```diff
@@ -12,6 +12,7 @@
 		"psycopg2",
 		"python-dateutil",
 		"requests",
+        "tzinfo",
```
ekimekim: inconsistent indentation, this file uses tabs

common/common/shifts.py
```diff
@@ -0,0 +1,83 @@
+
```
ekimekim, line 1: Applies to this file as a whole: tabs not spaces please! For consistency with the rest of the project if nothing else.

sheetsync/sheetsync/main.py
```diff
@@ -413,7 +417,14 @@ class PlaylistsSync(SheetSync):
 			event_id: The id of the streamlog event to sync
 	""",
 )
-def main(dbconnect, sync_configs, metrics_port=8005, backdoor_port=0, media_dir="."):
+@argh.arg('--timezone', help="Local timezone for determining shift times")
```
ekimekim, line 420: Could we make the timezone part of the `shifts` json instead? It removes the possibility of giving one without the other, when they only make sense together. It can be another top-level key.
It would also mean one less variable to need to plumb through everything.

sheetsync/sheetsync/sheets.py
```diff
@@ -301,7 +306,7 @@ def parse_bustime(self, value, preserve_dash=False):
 			return None
 		if value.strip() == "--":
 			return "--" if preserve_dash else None
-		bustime = common.parse_bustime(value)
+		bustime = common.shifts.parse_bustime(value)
```
ekimekim, line 309: I don't think `parse_bustime` moved?

docker-compose.jsonnet
```diff
@@ -176,6 +176,27 @@
...
+                repeating: [
+                            ["Zeta Shift", 0, 6],
+   					        ["Dawn Guard", 6, 12],
+                            ["Alpha Flight", 12, 18],
+							["Night Watch", 18, 24],
+                           ],
+			    one_off: [
+                          ["Tech Test", null, $.bustime_start],
+				    	  ["Omega Shift", "http://example.com/omega_start.html", null],
+                         ]
```
ekimekim, line 196: not sure what's going on with the indentation here.

docker-compose.jsonnet
```diff
@@ -536,6 +557,8 @@
       image: $.get_image("sheetsync"),
       // Args for the sheetsync
       command: [
+		"--timezone", $.timezone,
```
ekimekim, line 560: inconsistent indentation - this file uses 2 spaces.

ekimekim (commented):
I'm happy with the general approach. Comments are mostly nitpicks, occasional bugs/desired changes in behaviour but nothing requiring a massive overhaul.

common/common/shifts.py
```diff
@@ -0,0 +1,83 @@
...
+    """Parse a shifts definition
+
+       The shifts definition is two entry mappable with two keys, repeating and one-off.
+
+       The repeating shifts entry is a list of shift definition. Each of these is a sequence consisting of the name of the shift, the starting hour of the shift in local time, and the ending hour in local time. Repeating shifts extending across midnight can be handled by using two shifts with the same name. For example:
+        [['Night', 0, 6],
+         ['Day', 6, 18],
+         ['Night', 18, 24]]
+
+       The one-off shifts entry is a list of shift definitions. Each of these is a sequence consisting of the name of the shift, the start the shift, and the end of the shift. A start or end time can be a timestamp, a URL or None. If it is a URL, the URL will be queried for a timestamp. If no timezone info is provided the timestamp will be assumed to be UTC. If the start time is None, then the start will be assumed to be the earliest possible datetime; if the end is None, it will be assumed to be the oldest possible datetime. If both the start and end are None, the shift will be ignored. For example:
```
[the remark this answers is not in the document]
chrusher: Having both None be a 10 000 year long shift makes the logic a bit simpler

common/common/shifts.py
```diff
@@ -0,0 +1,83 @@
...
+def calculate_shift(time, shifts, timezone):
+    """Calculate what shift a time falls in. 
+        
+       time is a datetime, shifts the output from parse_shifts and timezone a 
+    """
+    if not time:
+        return ''
+    
+    for shift in shifts['one-off']:
+        print(time, shift[1], shift[2])
```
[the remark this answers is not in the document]
chrusher: Doh!

common/common/shifts.py
```diff
@@ -0,0 +1,83 @@
...
+        if shift[1] <= time < shift[2]:
+            return shift[0]
+        
+    #since shifts are based on local times we have to worry about timezones for once
+    local_time = time.replace(tzinfo=UTC).astimezone(timezone)
+    # do a more involved calculation to allow for non-integer start and end hours
+    time_diff = local_time - datetime.datetime(local_time.year, local_time.month, local_time.day, tzinfo=timezone)
+    hour = time_diff / datetime.timedelta(hours=1)
```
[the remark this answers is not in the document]
chrusher: Fucking DST, making everything more complicated than it should be.

thrimshim/thrimshim/main.py
chrusher: I think my new solution of looping over the tag list of each playlist when creating the `default_templates` dictionary gracefully handles both playlists without any tags and ones with multiple tags.

sheetsync/sheetsync/main.py
chrusher: Good idea.

chrusher:
I believe I have addressed all of your comments ekim. I added a bit more code to `sheetsync` to make things more robust if no shift definitions are provided or there are errors parsing the shift definition.

chrusher:
I have now tested the `thrimshim` part of the changes and the`thrimshim` can correctly select templates using tags and the `playlists` table.

sheetsync/sheetsync/sheets.py
```diff
@@ -273,6 +274,9 @@ def __init__(self, client, sheet_id, worksheets, bustime_start, edit_url, alloca
 		self.bustime_start = bustime_start
 		self.edit_url = edit_url
 
+		# fallback to no shifts if there is a shift parsing error
+		self.latest_shifts = {'repeating':[], 'one_off':[], 'timezone':zoneinfo.ZoneInfo('UTC')}
```
ekimekim, line 278: I think it would be better to fail entirely (and crash) if the initial parse fails, since this more likely represents a config error. And we don't want a misconfiguration to result in changing ALL videos to remove existing shifts.
```suggestion
		self.latest_shifts = common.shifts.parse_shifts(self.shifts)
```

sheetsync/sheetsync/sheets.py
```diff
@@ -294,7 +298,11 @@ def __init__(self, client, sheet_id, worksheets, bustime_start, edit_url, alloca
 
 	def get_rows(self):
 		# only need to update the shifts once per sync
-		self.latest_shifts = common.shifts.parse_shifts(self.shifts)
+		try:
+			self.latest_shifts = common.shifts.parse_shifts(self.shifts)
+		except Exception as e:
+			logging.error('Error parsing shifts with {}. Using previous shifts definition.'.format(e))
```
ekimekim, line 304: nitpick: Instead of printing the error explicitly, use `logging.exception` which automatically includes the error and traceback.
```suggestion
		except Exception:
			logging.exception('Error parsing shifts. Using previous shifts definition.')
```

sheetsync/sheetsync/sheets.py
```diff
@@ -331,9 +339,10 @@ def parse_row(self, worksheet, row_index, row):
 		# This is only needed for full events (not the archive sheet),
 		# so only do it if we had a tags column in the first place.
 		if 'tags' in row_dict:
+			shift_tag = common.shifts.calculate_shift(row_dict['event_start'], self.current_shifts)
```
ekimekim, line 342: typo: The variable is called `self.latest_shifts` above but `self.current_shifts` here.
```suggestion
			shift_tag = common.shifts.calculate_shift(row_dict['event_start'], self.latest_shifts)
```

thrimshim/thrimshim/main.py
```diff
@@ -242,22 +242,25 @@ def get_row(ident):
...
+				WHERE default_template IS NOT NULL
+			"""
 			results = database.query(conn, query)
 			default_templates = {}
 			for row in results:
-				row = row._asdict:
-				if row['name'] and row['default_template']:
-					default_templates[row['name']] = row['default_template']
+				for tag in row.tag:
+					default_templates[tag] = row.default_template
```
ekimekim: I don't think this is correct. Suppose you have a playlist with tags `foo` and `bar`. This code will use that playlist's default template if `foo` *or* `bar` is in the event tags, whereas the playlist only matches if `foo` *and* `bar` are in the event tags.
It would also break if there is one playlist that matches `foo` and another that matches `foo` and `bar`, ie. it assumes tags are unique.

The multi-tag stuff is confusing, I think the easiest option for now is to only consider single-tag playlists. But if we do  support multiple tags we should only match if the video would be added to the playlist. It's just a question of how that would interact with the tag order priority. Maybe take the matching playlist with the latest tag that is part of that playlist's tags? so eg. with tags `foo, bar, baz` on the event, and playlists `foo, baz` and `bar`, `foo, baz` would take priority because `baz` comes last.

ekimekim (approved):

PR #449, opened 2024-11-05
Sheetsync fixes

chrusher:


ElementalAlchemist:
After several passes, we have a sheetsync that syncs successfully again.

ElementalAlchemist (approved):

Issue #450, opened 2024-11-06
CI build speedups


HubbeKing:
From looking at logs, our current main culprits for slow CI builds are:
- `buscribe_api` builds `gevent` from source, which takes a couple minutes 
    - see https://github.com/dbvideostriketeam/wubloader/actions/runs/11693051936/job/32563745076#step:4:119
    - check timestamps by looking at the settings cogwheel and checking "show timestamps"
- `buscribe` downloads .zip files and extracts them, and then throws out the .zip files. this takes 5-6 minutes.
    - compare https://github.com/dbvideostriketeam/wubloader/actions/runs/11693051936/job/32563744792#step:4:1
    - compare https://github.com/dbvideostriketeam/wubloader/actions/runs/11697085596/job/32575244800#step:4:23
    - a build with step 3 cached is *much* faster
- Github Actions seems very fucky when it does and doesn't use the cache

The first two are probably reasonably easy to sort out. `buscribe_api` is missing `RUN pip install --upgrade wheel` from its Dockerfile, which will likely fix it so it can use the gevent wheel. The .zip files used by `buscribe` could probably be shoved into the repo? We'd have to look at the licenses for Vosk, since that's where the .zip files are from.

Github actions' cache fuckery is probably trickier, but fixing the first two should help loads already

HubbeKing:
I'll make a PR for the fixes after work, but if anyone else wants to do in the meantime that feel free

chrusher:
`buscribe` is also missing `RUN pip install --upgrade wheel` which sped up building other wubloader components in the past.

HubbeKing:
`buscribe` appears to already be using wheels correctly, so it's probably an issue with the alpine linux images that is currently only impacting `buscribe_api`
https://github.com/dbvideostriketeam/wubloader/actions/runs/11693051936/job/32563744792#step:4:2149

HubbeKing:
Thought about using Github's Action Cache for the .zip files, but it turns out they auto-evict caches that haven't been accessed in 7 days, so that's not very useful. Will look at just shoving them into the repo.

HubbeKing:
Probably not worth looking at more speedups, GH Actions currently builds the whole project in a couple minutes at most.

PR #452, opened 2024-11-06
Fixing pip in buscripe_api

chrusher:
A small change to speed up installing Python packages in `buscripe_api`

Installing wheels works fine in `buscripe` since it is using a Debian base image.

Partial fix for #450


HubbeKing (approved):
Looks like it successfully installed `gevent` by downloading the wheel.
https://github.com/dbvideostriketeam/wubloader/actions/runs/11699323453/job/32581104944?pr=452#step:4:28

LGTM!

PR #454, opened 2024-11-06
Pass the auth token when submitting thumbnails

ElementalAlchemist:
Simply, this ensures we pass the auth token when submitting thumbnails.


chrusher (approved):
Looks good to me

Issue #399, opened 2024-08-10
Warn if not showing chat in stream page due to no end time


ekimekim:
Easier stopgap to https://github.com/dbvideostriketeam/wubloader/issues/398

The current behaviour is confusing as it's not clear why the chat doesn't show up.
This can simply be a static message where chat would go saying "set an end time to get chat" or similar.

Issue #398, opened 2024-08-10
Show chat in stream page even if end time not set


ekimekim:
Currently we only show chat when both start and end are known.

PR #456, opened 2024-11-07
thrimbletrimmer: When no end time set, show and auto-refresh chat

ekimekim:
We request current time + 1 minute, and re-request every 10 seconds.

If I'm reading the code correctly, everything already correctly handles the chat being replaced
by a new fetch (as though we loaded a new video range).

### DIFF
--- a/thrimbletrimmer/scripts/common.js
+++ b/thrimbletrimmer/scripts/common.js
@@ -517,9 +517,12 @@ function triggerDownload(url, filename) {
 
 function sendChatLogLoadData() {
 	let startTime = getStartTime();
-	let endTime = getEndTime();
-	if (!startTime || !endTime) {
+	if (startTime === null) {
 		return;
+	}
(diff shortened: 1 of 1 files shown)


ElementalAlchemist:
Chat can sometimes take a really long time to render (this is why we put it off to a background worker). How long of ranges/much chat was this tested on? Should we do this less frequently than 10 seconds?

ekimekim:
This was not tested.

ekimekim:
My hope was that it would be frequent enough for someone watching live. If that's not feasible without a smart "what changed" algorithm, we can do without.

ElementalAlchemist:
I just remembered this also:
Doesn't the server-side chat functionality only update with new chat once per minute?

ekimekim:
I've downgraded this PR to only fetch chat once, since a live view is not viable.

ekimekim:
Closes #398 
Closes #399

PR #459, opened 2024-11-07
Edit thumbnails through Thrimbletrimmer

ElementalAlchemist:
Adds a basic, incredibly not-fancy UI for editing thumbnail templates on the Thrimbletrimmer thumbnail manager page.

Fixes a typo in thrimshim related to updating templates.
Removes the duplicate name validation from the `validate-template` endpoint, as it simply requires a name change for all edits to thumbnails.

### DIFF
--- a/thrimbletrimmer/scripts/thumbnails.js
+++ b/thrimbletrimmer/scripts/thumbnails.js
@@ -126,35 +126,302 @@ window.addEventListener("DOMContentLoaded", async (event) => {
 	}
 });
 
-function addTemplate(index, template) {
+function generateTemplateDOM(index, template) {
 	const { name, description, attribution, crop, location } = template;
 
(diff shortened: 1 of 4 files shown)


chrusher (commented):
I left out an `if` statement when I was writing this

What I meant to write is this 

```	
       # check if new name is in database
       if name != new_template['name']:
		query = sql.SQL("""
			SELECT name FROM templates WHERE name = %s 
		""")
		results = database.query(conn, query, new_template['name'])
		if results.fetchone() is not None:
			return 'Template with name {} already exists'.format(new_template['name']), 400
			```

ElementalAlchemist:
You're right, that's a much better idea.

chrusher (approved):
Looks good to me now

PR #461, opened 2024-11-08
thrimshim: use send_file to serve templates from database

### LINKED ISSUE #458: Investigate performance of downloading thumbnail template images
Probably a "next year" at this point, but it takes me about 6 seconds to download a 100kB thumbnail template image like https://toodles.videostrike.team:1337/thrimshim/template/tech_test.png, and there are no response headers enabling client-side caching.

Oddly, downloads of a built thumbnail are fast(er). Accessing /thumbnail/desertbus/source.png takes one second for a composited image, when the template alone is extremely slow. Even if the issue is that the raw template data is stored as a database blob, there seems to be a problem here - if it takes 6 seconds to get the template from the database, it should take more than 6 seconds to serve a composite image that depends on that template.

I suspect that there's some bottleneck in the thrimshim/template image serving process. Is it normal to use `flask.Response` rather than `send_file`? https://github.com/dbvideostriketeam/wubloader/blob/master/thrimshim/thrimshim/main.py#L631

Even if there isn't, we could get significant UX gains by allowing some level of client-side caching.

dcollinsn:
I don't claim to understand why this works, but this makes downloading the thumbnail templates (to preview on the thumbnail management page, or to select the "hole" in the advanced crop settings in the editor) way faster. Something to do with how flask is chunking the `memoryview` object to serve.

Applying this changed the timings to download "fiddling.png" from 818ms waiting and 6770ms receiving, to 777ms waiting and 14ms receiving.

Resolves #458

### DIFF
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -1,5 +1,6 @@
 import datetime
 from functools import wraps
+import io
 import json
 import logging
 import re
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -628,7 +628,12 @@ def get_template(name):
 
 		image = row[0]
 		logging.info('Thumbnail image of {} fetched'.format(name))
-		return flask.Response(image, mimetype='image/png')
+		return flask.send_file(
+			io.BytesIO(image),
+			mimetype='image/png',
+			as_attachment=False,
+			download_name=f'{name}.png'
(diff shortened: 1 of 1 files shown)


chrusher:
It will not spot all Python bugs but running a linter such as Pyflakes before pushing is good practice.

dcollinsn:
ekim has a better fix (`bytes(image)` apparently is enough), closing

PR #462, opened 2024-11-08
Fix uploading a new thumbnail image when updating a thumbnail

ElementalAlchemist:
I forgot to update this spot or something.


chrusher (approved):
I trust you are now performing the correct JavaScript incantations

PR #464, opened 2024-11-08
Set up multiplatform builds

HubbeKing:
With QEMU we should be able to do multiplatform builds in Github Actions. Gonna see about setting that up.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,7 +41,7 @@ jobs:
           - zulip_bots
     steps:
       - name: Check out repo
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           submodules: recursive
           # By default, for PR builds, Actions will check out a merge commit between the actual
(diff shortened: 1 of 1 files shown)


HubbeKing:
aaw beans. something somewhere pulls in `libc-bin`... of course that's not gonna work.

chrusher:
How complicated is it to not build buscribe for arm64 but to build it for amd64?

HubbeKing:
Should be pretty easy to just skip arm64 for buscribe in the build script

HubbeKing:
or, y'know, it would be easy if I could type

HubbeKing:
well, I'm doing *something* wrong, but I'm unsure what. Will look more later.
`Error: pushing image "ghcr.io/dbvideostriketeam/wubloader-postgres:3ffab22" to "docker://ghcr.io/dbvideostriketeam/wubloader-postgres:3ffab22": no image found in manifest list for architecture amd64, variant "", OS linux: image not known`

HubbeKing:
I suspect the caching might be tripping it up? Buildah has some fixed issues about the `--layers` arg causing similar errors in other tasks, so it could be that... am reluctant to disable it though, we only just got it working again =/

HubbeKing:
grumble. i thought updating buildah might help, but the ubuntu github actions running has the official ubuntu apt repos, and they don't have anything newer than 1.33.7. Which is fun when the latest version is 1.37.5

HubbeKing:
and github doesn't want to provide other distros as runners. great. guess I have to look into how to make the build step use a docker image instead. or we could self-host a debian runner.

HubbeKing:
So, we currently cannot easily do this - we need a more up-to-date version of `buildah` in the CI tool, and Github's hosted runner does not make that like, possible.

We either:
- wait until Ubuntu 26.04 releases, and use that runner.
- self-host a runner on a not-stupid OS, like debian 12 or something.
- use a different CI tool.

PR #465, opened 2024-11-08
An attempt at rewriting the chat merge operaiton to be faster for large batches

ekimekim:
Not a big problem currently, though would be nice to load chat in restreamer faster.
This was just an idea on an alternate approach.

Current status: It's about 2x faster, but it produces different results. Might be entirely broken. Need to dig more when I have time to come back to it.

### DIFF
--- a/merge_v1.py
+++ b/merge_v1.py
@@ -1,0 +1,96 @@
+import json
+import time
+import hashlib
+
+# Calculates intersection of time range of both messages, or None if they don't overlap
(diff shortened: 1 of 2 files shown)


ekimekim:
After two more iterations of algorithms, I now have something which is optimized for fast performance when merging large collections of mostly non-overlapping batches (ie. the "fetch all chat in this time range" operation) while still being about the same performance when merging mostly-identical batches.

PR #466, opened 2024-11-08
thrimshim: Fix performance issue with getting template png

ekimekim:
Bytes returned from the database are a "memoryview" and not a bytes object.
These mostly behave the same, but one difference is that Flask recognizes
a bytes object as something it can send as-is, whereas a memoryview ends up using
its generic "iterable" processing. This results in sending every individual byte of the result
as a single part of a HTTP chunked encoding response, adding 5 bytes and a syscall to every byte returned.

The solution is to explicitly convert it to bytes before returning.


chrusher:
We should also update `thrimshim.get_template` since it is also pulling an image from the database.

PR #467, opened 2024-11-08
Update bus_analyzer to work with 2024 buscam

ekimekim:
- Updated pixel values
- New prototypes for main digits
- Cropped and resized prototypes for last odo digit - a lot of work to redo these and this is probably good enough.
- In basic testing, it seems to have trouble with 3 vs 6 vs 9, but otherwise seems to work.


chrusher (approved):
Looks good to me.

chrusher:
Between this and #463 `bus_analyzer` should work as well as last year.

PR #468, opened 2024-11-08
pubbot: Get channel ids from config and handle prizes in the new website

ekimekim:
Expects a hard-coded list of prize ids. We could in theory find open prize ids by parsing the html of the silent auctions page
but I'd rather not do that if we don't have to.

No longer displays the prize title in Zulip for the same reason, it just shows the id and links to it.


chrusher (approved):
Looks ok for me

PR #472, opened 2024-11-09
Add logging call for VOSK in-progress, and move to f-string syntax

MC42:


HubbeKing:
They'll all fail yeah, since you're a first-time contributor without write access. surprised they tried to run at all, honestly

buscribe/buscribe/buscribe.py
```diff
@@ -55,6 +55,8 @@ def transcribe_segments(segments: list, sample_rate: int, recognizer: BuscribeRe
 
                 if line_start_time > start_of_transcript:
                     write_line(result_json, line_start_time, line_end_time, db_cursor)
+                    logging.info(f"Line successfully written for duration {line_end_time - line_start_time}")
```
ekimekim, line 58: I'm much more interested in the actual start time, rather than the duration.
```suggestion
                    logging.info(f"Line successfully written for {line_start_time} to {line_end_time}")
```

ekimekim (approved):
I triggered the container builds. They only do a build (to make sure it works), they can't affect anything else.

ekimekim:
Sigh right it shouldn't be trying to push if it doesn't have write perms. TODO fix that.

HubbeKing (approved):

ekimekim (approved):

PR #495, opened 2024-11-19
Install/readme.md tidy up and gitignore additions.

MC42:
While incomplete, it's a small tidy-up on the configs (and the gitignore should now make it harder to commit credentials.)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,14 +19,9 @@ This installation guide is written assuming you are on a Linux-like operating sy
   
 ## Download the wubloader
 
-You can download the latest version of the wubloader from github:
-
-  https://github.com/dbvideostriketeam/wubloader/archive/master.zip
-  
-Alternatively if you have `git` installed you can clone the git repository:
+Use github to clone the repository into a working directory.  If you intend to make changes, it's highly suggested you clone it into a fork of the repo on your own GitHub account.
 
   `git clone https://github.com/dbvideostriketeam/wubloader`
-  
   
 ## Generate the docker-compose file
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -51,7 +51,7 @@ By default the `downloader`, `restreamer`, `backfiller`, `cutter`, `thrimshim`, 
 
 If you are running a `cutter` you will have to place the appropriate Google credentials in a JSON file given by the `cutter_creds_file`. Likewise, if you are running the `sheetsync` service, you will have to place the appropriate credentials in the JSON file pointed to by `sheetsync_creds_file` as well as set the appropriate `sheet_id` and `worksheets` for the Google sheet to sync with. You will also need to set the appropriate `edit_url` to access `thrimbletrimmer`.  
 
-## Running the wubloader
+## Running Wubloader
 
 To start the wubloader, simply run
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -7,23 +7,37 @@ as part of [Desert Bus For Hope](https://desertbus.org).
...
   ie. it replicates missing segments.
-* `cutter` interacts with a database to perform cutting jobs
+* `chat_archiver` records twitch chat messages and merges them with records from other nodes.
+
+#### Processing
 * `sheetsync` syncs specifc database columns to a google doc which is the primary operator interface.
+* `cutter` interacts with a database to perform cutting jobs
 * `thrimshim` acts as an interface between the `thrimbletrimmer` editor and the database.
 * `thrimbletrimmer` is a browser based video editor.
 * `segment_coverage` regularly checks whether there is complete segment coverage for each hour.
```
chrusher, line 26: I would include segment_coverage under Ingest since it is something run on each of the replication nodes

README.md
```diff
@@ -7,23 +7,37 @@ as part of [Desert Bus For Hope](https://desertbus.org).
...
-* `chat_archiver` records twitch chat messages and merges them with records from other nodes.
-* `database` hosts a Postgres database to store events to be edited.
+
+#### Analysis
+* `bus_analyzer` does OCR on the stream to produce timestamps of/for progress through the route.
+* `buscribe` is back-end audio speech-to-text extraction.
+* `buscribe_api` is the API provided for below to access the stored results of transcription.
+* `buscribe-web` is the web frontend to interface with `buscribe`'s output (via `buscribe_api`)
+
+#### Services
```
chrusher, line 35: I would include the Zulip bots here.

INSTALL.md
```diff
@@ -49,9 +44,9 @@ After making any changes to `docker-compose.jsonnet`, you will need to rerun `ge
 
 By default the `downloader`, `restreamer`, `backfiller`, `cutter`, `thrimshim`, `segment_coverage` and `nginx` services of the wubloader will be run. To change which services are run edit the `enabled` object in `docker-compose.jsonnet`. A complete wubloader set up also requires one and only one `database` service (though having a backup database is a good idea), one and only one `sheetsync` service and one and only one `playlist_manager` service.
```
chrusher, line 45: Change `database` to `postgres`

README.md
```diff
@@ -7,23 +7,37 @@ as part of [Desert Bus For Hope](https://desertbus.org).
 A full design doc can be read at [initial-design-doc.pdf](./initial-design-doc.pdf),
 but a brief overview of the components:
 
+#### Shared Components
+
```
chrusher, line 11: I would move ngnix here since it is in front of everything

ekimekim: I think the current taxonomy is fine, with "shared" literally meaning "the same code is used in multiple services". We have the "Services" section for things that everything uses (nginx, postgres).

README.md
ekimekim: segment_coverage probably belongs under analysis, as it doesn't involve actually grabbing data from somewhere.

Or the other option is to completely restructure this section into "replication nodes", "edit nodes" etc. Which may be more useful.

ekimekim (approved):
This all seems reasonable and good cleanup. Thanks!

Issue #479, opened 2024-11-10
Thrimbletrimmer proposal: Guidelines or templates by event type


dcollinsn:
"next year" idea:

Giveaways in particular require some specific changes to be made by the editor, and I don't always remember what they are: Remove the donation amount, put the total raised in the description only, and thank the donor/crafter. There may be other tips like that for many event types, known either before the run, or, developed during a run as we learn how to edit/upload/thumbnail new gags.

Should we have a way to provide some textbox that appears on the editor for a specific event type, so that editors can be reminded of this kind of thing?

ekimekim:
This ties into https://github.com/dbvideostriketeam/wubloader/issues/504 which wants RDP-specific behaviour

ekimekim:
Can use `notes` field for this, have thrimshim automatically insert based on event type.

PR #522, opened 2025-01-19
Thrimbletrimmer: Default video title to blank

### LINKED ISSUE #504: Thrimbletrimmer should not set a default video title
Currently, thrimbletrimmer will fill in both the video title and description using the event description text.
However, what makes for a good sheet description is not always a good video title.
This default leads to a large number of videos with unmodified titles being uploaded.

We want to change the default title to be empty (which is an error to submit), forcing the editor to think about it.
They may still wish to copy-paste the description from the description box to the title box if it really is a good title,
but this change means they need to make a concious decision to do that instead of it being the default.

dcollinsn:
To encourage editors to write their own.

This patch simply uses the RDP tag to decide whether to use the row description as a default for the video title, or for the video description. This will probably be changed by #479 down the line, but that's lower priority and more complex, so this is a good starting point.

Resolves #504

### DIFF
--- a/thrimbletrimmer/scripts/edit.js
+++ b/thrimbletrimmer/scripts/edit.js
@@ -655,7 +655,9 @@ async function initializeVideoInfo() {
 	titlePrefixElem.innerText = videoInfo.title_prefix;
 
 	const titleElem = document.getElementById("video-info-title");
-	if (videoInfo.video_title) {
+	if (videoInfo.video_title !== null) {
+		// If a video titles was saved (even if it is blank), use that. Titles
+		// can't currently be blank, but we may be loosening validation for drafts.
 		titleElem.value = videoInfo.video_title;
 	} else {
 		// If a video title hasn't been set yet, leave it blank.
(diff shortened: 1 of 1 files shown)


ekimekim (approved):

ElementalAlchemist:
This would have been easy to do in thrimbletrimmer... if drafts didn't exist. Which is why I didn't.
If someone saves a draft (which has a legitimately useful title) and then loads the video again, does this just ignore the title thrimshim sends from the database? (Same with the description for videos with RDP in the title)

ekimekim:
If a draft has never been saved, `video_title` and `video_description` are null. If it has, they'll be a string (possibly an empty one for description).
Apart from the latter case (an intentionally empty description) the current logic (use `video_title`/`video_description` if truthy, otherwise use the defaults) should do what we want.

thrimbletrimmer/scripts/edit.js
```diff
@@ -668,7 +672,11 @@ async function initializeVideoInfo() {
 	if (videoInfo.video_description) {
```
ekimekim, line 672: If `video_description` is the empty string (ie. someone deleted the description then saved draft) we don't want to use the default. We only want that when `video_description` is `null`:
```suggestion
	if (videoInfo.video_description === null) {
```

dcollinsn: With how the logic is currently organized, that should be `!== null`, right?

ekimekim (approved):

PR #523, opened 2025-01-20
Preview transitions in a popup

### LINKED ISSUE #507: Thrimbletrimmer: Transition previews
It is frequently unclear how a transition is going to look when setting it up in the editor. We've had to frequently re-edit videos due to transitions needing to be tweaked.
We should have a "Preview" button next to each transition which links to a restreamer cut link showing:
- A few seconds before the transition start
- The transition
- A few seconds after the transition end

This should really be in a format that users can load directly in their browser, instead of a file they have to download.

We should also clarify that transition timing is that the duration is the *overlap* between the two ranges, and the ranges still start/end where indicated.

dcollinsn:
This patch allows previewing a video transition. It forces a 480p popup and loads a `<video>` element containing a full cut webp starting 5 seconds before the transition begins, and ending 5 seconds after it ends.

Resolves #507

### DIFF
--- a/restreamer/restreamer/main.py
+++ b/restreamer/restreamer/main.py
@@ -411,6 +411,11 @@ def cut(channel, quality):
 		stream, muxer, mimetype = (True, 'mpegts', 'video/MP2T') if type == 'mpegts' else (False, 'mp4', 'video/mp4')
 		encoding_args = ['-c:v', 'libx264', '-preset', 'ultrafast', '-crf', '0', '-f', muxer]
 		return Response(full_cut_segments(segment_ranges, ranges, transitions, encoding_args, stream=stream), mimetype=mimetype)
+	elif type == 'webm':
+		# Basic webm settings to work in Firefox and Chrome.
+		stream, muxer, mimetype = (True, 'webm', 'video/webm')
+		encoding_args = ['-c:v', 'libvpx-vp9', '-c:a', 'libopus', '-f', muxer]
+		return Response(full_cut_segments(segment_ranges, ranges, transitions, encoding_args, stream=stream), mimetype=mimetype)
 	else:
 		return "Unknown type {!r}".format(type), 400
 
--- a/restreamer/restreamer/main.py
+++ b/restreamer/restreamer/main.py
@@ -413,9 +413,8 @@ def cut(channel, quality):
 		return Response(full_cut_segments(segment_ranges, ranges, transitions, encoding_args, stream=stream), mimetype=mimetype)
 	elif type == 'webm':
 		# Basic webm settings to work in Firefox and Chrome.
-		stream, muxer, mimetype = (True, 'webm', 'video/webm')
-		encoding_args = ['-c:v', 'libvpx-vp9', '-c:a', 'libopus', '-f', muxer]
-		return Response(full_cut_segments(segment_ranges, ranges, transitions, encoding_args, stream=stream), mimetype=mimetype)
+		encoding_args = ['-c:v', 'libvpx-vp9', '-c:a', 'libopus', '-f', 'webm']
+		return Response(full_cut_segments(segment_ranges, ranges, transitions, encoding_args, stream=True), mimetype='video/webm')
 	else:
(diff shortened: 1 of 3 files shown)


dcollinsn:
Not sure if there's anything else that you want in the popup, or if just the video element is okay. There is a bit of a delay - 5 to 10 seconds - before the video starts playing, I tried a few additional flags like `-deadline good -cpu-used 5` and it didn't noticably speed anything up, but I haven't benchmarked.

I didn't add the webp option to the types of normal downloads because of server load, similar to the mp4 transcode.

restreamer/restreamer/main.py
```diff
@@ -411,6 +411,11 @@ def cut(channel, quality):
 		stream, muxer, mimetype = (True, 'mpegts', 'video/MP2T') if type == 'mpegts' else (False, 'mp4', 'video/mp4')
 		encoding_args = ['-c:v', 'libx264', '-preset', 'ultrafast', '-crf', '0', '-f', muxer]
 		return Response(full_cut_segments(segment_ranges, ranges, transitions, encoding_args, stream=stream), mimetype=mimetype)
+	elif type == 'webm':
+		# Basic webm settings to work in Firefox and Chrome.
+		stream, muxer, mimetype = (True, 'webm', 'video/webm')
+		encoding_args = ['-c:v', 'libvpx-vp9', '-c:a', 'libopus', '-f', muxer]
+		return Response(full_cut_segments(segment_ranges, ranges, transitions, encoding_args, stream=stream), mimetype=mimetype)
```
ekimekim, line 418: nitpick: we aren't tweaking the options in two different ways like on the `mpegts`/`mp4` path, so we can just inline the `stream`, `muxer` and `mimetype` values:
```suggestion
		encoding_args = ['-c:v', 'libvpx-vp9', '-c:a', 'libopus', '-f', 'webm']
		return Response(full_cut_segments(segment_ranges, ranges, transitions, encoding_args, stream=True), mimetype='video/webm')
```

thrimbletrimmer/scripts/edit.js
```diff
@@ -1655,6 +1655,43 @@ function updateDownloadLink() {
...
+		if (previewTimeRanges[0].start && previewTimeRanges[0].end && previewTimeRanges[1].start && previewTimeRanges[1].end) {
+			// Adjust segment start and end times based on calculated padding.
+			previewTimeRanges[0].start = wubloaderTimeFromDateTime(luxon.DateTime.max(
+				dateTimeFromWubloaderTime(previewTimeRanges[0].start),
+				dateTimeFromWubloaderTime(previewTimeRanges[0].end).minus(luxon.Duration.fromMillis(thisPaddingDurationSeconds*1000))
+			));
+			previewTimeRanges[1].end = wubloaderTimeFromDateTime(luxon.DateTime.min(
+				dateTimeFromWubloaderTime(previewTimeRanges[1].end),
+				dateTimeFromWubloaderTime(previewTimeRanges[1].start).plus(luxon.Duration.fromMillis(thisPaddingDurationSeconds*1000))
+			));
```
ekimekim, line 1681: I don't think mutating these `previewTimeRanges` objects is safe. `timeRanges.slice()` on line 1664 won't make a copy of the time range object, it will take a reference. We then mutate those objects, in particular shortening the second one's `end` field.
Then in the next iteration we grab the same one again (but now it's the first of the pair in the slice) and use its `end` (minus the padding) to set the start. So if you have three ranges and two transitions, this would result in the preview for the second transition covering the entire middle range, not just the padding.

The fix would be to make new range objects to pass into `generateDownloadURL` here instead of mutating:
```
previewTimeRanges = [
  {
    start: ... // as current,
    end: previewTimeRanges[0].end,
  },
  {
    start: previewTimeRanges[1].start,
    end: ... // as current
  },
];
```

ekimekim (approved):
One bug (unless I've misunderstood something), apart from that LGTM.

ekimekim:
continued in https://github.com/dbvideostriketeam/wubloader/pull/549

PR #524, opened 2025-02-19
Updated Postgres version

### LINKED ISSUE #451: Postgres upgrade
We should upgrade from Postgres 12 to a more recent version, 12 is getting its final release on November 14th 2024 and is EOL after that.
- https://github.com/dbvideostriketeam/wubloader/blob/master/postgres/Dockerfile
- https://www.postgresql.org/support/versioning/
- https://endoflife.date/postgresql
- https://www.postgresql.org/docs/current/upgrading.html

13 would of course work, but we should probably do this after DB and just go right up to 16 or 17.

chrusher:
Updated the version of Postgres we use to 17. This is not been tested. Will fix #451

### DIFF
--- a/postgres/Dockerfile
+++ b/postgres/Dockerfile
@@ -1,4 +1,4 @@
-FROM postgres:12
+FROM postgres:17
 COPY postgres/setup.sh /docker-entrypoint-initdb.d/setup.sh
 COPY postgres/schema.sql /
 COPY postgres/buscribe.sql /
--- a/postgres/setup.sh
+++ b/postgres/setup.sh
@@ -1,6 +1,6 @@
 #! /bin/bash
 
-set -e
+set -eu
 
 sql() {
 	local user
@@ -14,9 +14,10 @@ sed -i "/host all all all/d" "$PGDATA/pg_hba.conf"
 echo "host all $WUBLOADER_USER all md5" >> "$PGDATA/pg_hba.conf"
 
 echo "Creating $WUBLOADER_USER"
-sql "$POSTGRES_USER" <<-EOSQL
+sql "$POSTGRES_USER" -d "$POSTGRES_DB" <<-EOSQL
 
 CREATE USER $WUBLOADER_USER LOGIN PASSWORD '$WUBLOADER_PASSWORD';
+GRANT CREATE ON SCHEMA public TO $WUBLOADER_USER;
 
 EOSQL
 
@@ -32,11 +33,7 @@ if [ -n "$REPLICATION_USER" ]; then
 	EOSQL
 
 	cat >> ${PGDATA}/postgresql.conf <<-EOF
-	wal_level = replica
-	archive_mode = on
-	archive_command = 'cd .'
-	max_wal_senders = 8
-	wal_keep_segments = 8
+	wal_keep_size = 128MB
 	EOF
 
 fi


HubbeKing:
Should be fine, AFAIK we don't do any weird stuff that'd be dependent on old postgres, so it ought to Just Work™️

ekimekim:
Fails with:
```
Applying schema for wubloader
2025-02-19 13:13:33.131 UTC [67] ERROR:  permission denied for schema public
2025-02-19 13:13:33.131 UTC [67] STATEMENT:  CREATE TYPE event_state as ENUM (
		'UNEDITED',
		'EDITED',
		'CLAIMED',
		'FINALIZING',
		'TRANSCODING',
		'DONE',
		'MODIFIED'
	);
ERROR:  permission denied for schema public
```
This is due to the wubloader user not having permissions to create types (or tables, etc).
I'm not sure why this worked before?

The fix:
```diff
--- a/postgres/setup.sh
+++ b/postgres/setup.sh
@@ -14,9 +14,10 @@ sed -i "/host all all all/d" "$PGDATA/pg_hba.conf"
 echo "host all $WUBLOADER_USER all md5" >> "$PGDATA/pg_hba.conf"
 
 echo "Creating $WUBLOADER_USER"
-sql "$POSTGRES_USER" <<-EOSQL
+sql "$POSTGRES_USER" -d "$POSTGRES_DB" <<-EOSQL
 
 CREATE USER $WUBLOADER_USER LOGIN PASSWORD '$WUBLOADER_PASSWORD';
+GRANT CREATE ON SCHEMA public TO $WUBLOADER_USER;
 
 EOSQL
```

ekimekim:
Now fails with:
```
2025-02-19 13:20:23.652 GMT [1] LOG:  unrecognized configuration parameter "wal_keep_segments" in file "/var/lib/postgresql/data/postgresql.conf" line 849
2025-02-19 13:20:23.652 GMT [1] FATAL:  configuration file "/var/lib/postgresql/data/postgresql.conf" contains errors
```
This is due to `wal_keep_segments` being renamed `wal_keep_size`. While we're there, we can remove some superfluous options that enable WAL archiving but then use a no-op archive command (so it does nothing).
Also while we're there, `wal_level` now defaults to `replica` and `wal_max_senders` defaults to 10, which is no practical difference to our choice of 8 so we can remove it.
```diff
--- a/postgres/setup.sh
+++ b/postgres/setup.sh
@@ -32,11 +33,7 @@ if [ -n "$REPLICATION_USER" ]; then
        EOSQL
 
        cat >> ${PGDATA}/postgresql.conf <<-EOF
-       wal_level = replica
-       archive_mode = on
-       archive_command = 'cd .'
-       max_wal_senders = 8
-       wal_keep_segments = 8
+       wal_keep_size = 8
        EOF
 
 fi
```
As far as I can tell, our use of `wal_keep_size` is solely because of a race condition during standby startup between taking a basebackup and then beginning replication. When replication begins we need to catch up on any changes between taking the backup and now. Without `wal_keep_size`, those WAL segments would be unavailable. With `wal_keep_size` set to a fixed number, we have a small time window where it works but if it's too slow it won't. The proper fix would be to create a replication slot before starting the basebackup, which will hold all WAL segments while the basebackup is ongoing. But we can skip that for now.

ekimekim:
With those fixes, it works. At least enough for schemas to apply and for me to connect and query things.

ekimekim:
Merging since it's tested as best we can and I have other changes I want to put on top.

PR #525, opened 2025-05-23
Add youtubebot

ekimekim:
This adds a zulip bot that polls the youtube API for new comment threads, and posts them to Zulip.
Some limitations:
- It doesn't keep any state, so it won't post anything it "missed" while not running.
- It can only find top-level comments, not replies
- For quota reasons, we shouldn't poll more often than every 1 minute (at this rate we consume approx 1 upload worth of quota per day)
- If somehow there are more than 100 comments within 1 minute, it will miss all but the last 100.

### DIFF
--- a/docker-compose.jsonnet
+++ b/docker-compose.jsonnet
@@ -37,6 +37,7 @@
     pubbot: false,
     blogbot: false,
     prizebot: false,
+    youtubebot: false,
     bus_analyzer: false,
     graphs: false,
   },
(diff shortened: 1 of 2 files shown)


docker-compose.jsonnet
```diff
@@ -358,6 +359,13 @@
     state_path:: "./prizebot_state.json",
   },
 
+  youtubebot:: {
+    zulip_email: "youtube-bot@chat.videostrike.team",
+    zulip_api_key: "",
+    google_credentials_file: $.cutter_creds_file,
+    channel_id: "UCz5-PNQxaT4WtB_OMAwD85g", // DesertBusForHope
```
chrusher, line 366: I assume this is something publicly available.

ekimekim: yes, it's the channel id of the DB channel. anyone can look it up via the api based on the public channel name.

PR #526, opened 2025-05-27
Temporary debugging: Write out last 10 media playlists after fetching each segment

ekimekim:
This is some very old code that I've resurrected since we need it again. The PR is just so GHA builds images.

I'll clean this up into something we can keep merged but turned off later.


chrusher (approved):
Looks good to me

ekimekim:
This was added in 7b1c8c6d20eafffa111b840c395a8eb830b0e5e0

PR #556, opened 2025-10-25
Update the waveform if the live head isn't at the end time yet

### LINKED ISSUE #333: thrimbletrimmer - audio display is incorrect if row has an end time, but that end time + 2 minutes has not yet been reached
If a row has an end time, but the playlist corresponding to that row has not fully happened at the time the editor is opened (end time + 2 minutes), then the audio graph will still be loaded based on the segments that have been downloaded so far. The graph will be incorrect because it will span only the beginning of the playlist to the current time, while additional segments will be added to the video playlist as time progresses.

e.g.: start time is 100:01, end time is 100:03, and you load the editor at 100:03:30. The playlist *should* span 100:00:00 to 100:04:59.999. However, only segments from 100:00:00 to 100:03:30 actually exist.

The video editor initially spans from 100:00:00 to 100:03:30, but new segments will be added as they become available, until eventually the video editor spans 100:05:00.

The audio graph will be loaded when the editor is first loaded and is never updated, so it will always span 100:00:00 to 100:03:30.

Suggested fix: Any missing segments should be represented by a flatline in the audio graph, or alternately, if the playlist doesn't span the entire recommended range, the audio graph shouldn't be displayed at all.

ElementalAlchemist:
Fixes #333

### DIFF
--- a/thrimbletrimmer/scripts/edit.js
+++ b/thrimbletrimmer/scripts/edit.js
@@ -1003,11 +1003,14 @@ async function initializeVideoInfo() {
 function updateWaveform() {
 	const currentTime = DateTime.now().setZone("utc");
 
-	if (lastWaveformRequest && currentTime.diff(lastWaveformRequest).as("seconds") < 15) {
-		if (!waveformRequestTimeout) {
-			waveformRequestTimeout = setTimeout(updateWaveform, 15000);
+	if (lastWaveformRequest !== null) {
+	    const remaining = 15 - currentTime.diff(lastWaveformRequest).as("seconds");
+	    if (remaining > 0) {
+			if (waveformRequestTimeout === null) {
+				waveformRequestTimeout = setTimeout(updateWaveform, 1000 * remaining);
+			}
+			return;
 		}
-		return;
 	}
 
 	lastWaveformRequest = currentTime;


thrimbletrimmer/scripts/edit.js
```diff
@@ -1000,12 +1001,26 @@ async function initializeVideoInfo() {
...
 
 function updateWaveform() {
+	const currentTime = DateTime.now().setZone("utc");
+
+	if (lastWaveformRequest && currentTime.diff(lastWaveformRequest).as("seconds") < 15) {
+		if (!waveformRequestTimeout) {
+			waveformRequestTimeout = setTimeout(updateWaveform, 15000);
+		}
+		return;
+	}
```
ekimekim: nitpick: suppose an updateWaveform call comes in 14s after the previous. Then we will still wait a full 15s before executing it, for a total delay of 29s.
It would be more consistent to only delay for the difference in time, so the next update always happens 15s after the previous.


```suggestion
	if (lastWaveformRequest !== null) {
	    const remaining = 15 - currentTime.diff(lastWaveformRequest).as("seconds");
	    if (remaining > 0) {
			if (waveformRequestTimeout === null) {
				waveformRequestTimeout = setTimeout(updateWaveform, 1000 * remaining);
			}
			return;
		}
	}
```

Issue #497, opened 2025-01-14
Event audit log missing a lot of info


ekimekim:
Because of how it was written, the event audit log doesn't contain many relevant fields of the old row (or some relevant fields of the new row), which makes it very hard to reconstruct changes.

I suggest we stop trying to hand-roll this and replace it with a postgres trigger-based system, ideally something already implemented for us like https://wiki.postgresql.org/wiki/Audit_trigger_91plus

PR #560, opened 2025-10-27
replace events audit log and general postgres cleanup

ekimekim:
Uses https://wiki.postgresql.org/wiki/Audit_trigger_91plus to automatically track all changes to `events` and `templates` tables.

Changes schema setup so that the `postgres` user is the owner of all tables, and the `vst` user is explicitly granted permission to use them.

Misc script fixes.

Fixes https://github.com/dbvideostriketeam/wubloader/issues/497


chrusher (commented):
Looks ok to me

Issue #529, opened 2025-07-13
Add missing components to monitoring


chrusher:
The Zulip bots, and possibly other components, are not included in the monitoring dashboard. They should be added to the monitoring dashboard.

ekimekim:
Still do to:
- [ ] downloader: non-`full` segment downloads
- [ ] chat archiver: media download metrics, emote download metrics
- [ ] segment coverage image + link to full page
- [ ] playlist manager: general metrics, quota-using operation rates
- [ ] bus analyzer: up, general metrics
- [ ] buscribe: up, general metrics
- [ ] bots: up
- [ ] postgres: replication status

chrusher:
I will add monitoring to bus_analyzer

PR #562, opened 2025-10-28
zulip bots: basic metrics setup + running in no-zulip mode

### LINKED ISSUE #537: bots which save data should have redundancy
We have a variety of bots, some of which only post to zulip (like schedulebot), while others collect data and save it as well as posting it to zulip (like pubbot).

Of the ones that save data, we should run more than one copy, have each save data independently, but only have one post to zulip.

ekimekim:
Fixes #537

Part of #529

### DIFF
--- a/docker-compose.jsonnet
+++ b/docker-compose.jsonnet
@@ -289,6 +289,8 @@
   // Typically we set this to 1 month before bustime_start.
   buscribe_start:: "1970-01-01T00:00:00Z",
 
+  // Set this to empty string to disable sending to zulip in enabled bots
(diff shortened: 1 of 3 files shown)


ekimekim:
The `ports` setting specifies external (to the machine) port bindings, ie. "map this port on the host to this port on this container". Really we should only be defining them for nginx and postgres. Everything else was just for debugging (and should probably be removed now).
From one container to another (ie. nginx to the other service that is exposing metrics) we use the container's internal port, which is why the port mapping in the `nginx` section is independent of the main `ports` mapping.

PR #563, opened 2025-10-28
Add postgres exporter

ekimekim:
This third party exporter will collect a large range of useful stats by default,
most importantly replication lag.


chrusher (approved):
Looks ok to me

PR #564, opened 2025-10-29
metrics for missing components

### LINKED ISSUE #529: Add missing components to monitoring

ekimekim:
Closes #529, though we still need to add these to the dashboard.


chrusher (approved):
Looks ok to me

PR #565, opened 2025-11-01
prizebot: save data

### LINKED ISSUE #518: blog and prize info archiver
as distinct from zulip bots, this saves ALL info (both raw and as much as we can parse) for archival

ekimekim:
Fixes #518


chrusher (approved):
Looks ok to me

PR #570, opened 2025-11-04
driveclock: Specific point graphics for vegas vs tucson

ekimekim:
- Replaces the point graphic with two custom signposts for vegas and tucson
- Removes point graphic entirely in point progress mode - it doesn't look good and isn't needed


ElementalAlchemist:
Sorry, I didn't see this in time.

LGTM, as long as we're OK with point mode not showing the destination image.

Issue #561, opened 2025-10-27
Integrate post processing into bus_analyzer


chrusher:
The plan last year was to do post processing of the bus_analyzer output in a separate component, bus_synthesizer which would also act as the end point for getting the post processed readings. The plan now is for the results to be served by thrimshim and the post processing to be done within bus_analyzer.

PR #571, opened 2025-11-12
Adding post processing to bus_analyzer

chrusher:
Adds the post processing code that was in #483 into bus_analyzer solving #561 . Not yet tested.

### DIFF
--- a/bus_analyzer/bus_analyzer/main.py
+++ b/bus_analyzer/bus_analyzer/main.py
@@ -221,7 +221,7 @@ def analyze_hour(db_manager, prototypes, existing_segments, base_dir, channel, q
 
 	return [segment[1] for segment in segments_to_do]
 
-@timed(normalize=lambda result, db_manager, segments: len(segments))
+@timed(normalize=lambda result, db_manager, segments, channel: len(segments))
 def post_process(db_manager, segments, channel):
 
 	if segments:


bus_analyzer/bus_analyzer/main.py
```diff
@@ -218,6 +218,42 @@ def analyze_hour(db_manager, prototypes, existing_segments, base_dir, channel, q
 	for worker in workers:
 		worker.get() # re-raise errors
 
+	return [segment[1] for segment in segments_to_do]
+
+def post_process(db_manager, segments):
```
ekimekim: some quick and dirty monitoring:
```suggestion
@timed(normalize=lambda result, db_manager, segments: len(segments))
def post_process(db_manager, segments):
```

bus_analyzer/bus_analyzer/post_processing.py
```diff
@@ -0,0 +1,144 @@
...
+		next_diff = miles[i + 1] - miles[i]
+		if next_diff < 0 or next_diff > MAX_SPEED * (seconds[i + 1] - seconds[i]):
+			suspect.append(i)
+			continue
+		# handle big jumps to apparently good data
+		if good and miles[i] - miles[good[-1]] > MAX_SPEED * (seconds[i] - seconds[good[-1]]):
+			suspect.append(i)
+			continue
+		# try to filter out bad data at the start
+		if not good and miles[i] > 1000:
```
ekimekim: won't this prevent us from ever marking anything good when processing data that starts past 1000 miles (ie. halfway through point 3)?

bus_analyzer/bus_analyzer/main.py
```diff
@@ -220,7 +221,8 @@ def analyze_hour(db_manager, prototypes, existing_segments, base_dir, channel, q
 
 	return [segment[1] for segment in segments_to_do]
 
-def post_process(db_manager, segments):
+@timed(normalize=lambda result, db_manager, segments: len(segments))
```
ekimekim: since you're adding `channel` as an arg to the function it also needs to be in this lambda (it gets the same args as the function)
```suggestion
@timed(normalize=lambda result, db_manager, segments, channel: len(segments))
```

PR #598, opened 2026-03-04
Use debian:13 as a base for images rather than various versions of alpine

HubbeKing:
Debian images should give us:
- long-term support (debian:13 goes EOL in 2030)
- few changes in packages while we keep the same major version (Don't Break Debian)
- Hopefully faster builds due to more wheels and fewer compiles from source


HubbeKing:
Issues with `downloader` that need someone with more braincells:
- we install `deno` for yt-dlp js extraction. it's not in the debian repos, so we'll need to build from source with cargo. i'm assuming this is fine?
- alternatively we pull the latest release binary from github, but that means we only support `amd64`, which is probably also fine?
- we explicitly install `"prometheus-client==0.7.1", # locked version as we rely on internals` - is this still true?

api_ping/setup.py
```diff
@@ -5,7 +5,7 @@
 	version = "0.0.0",
 	packages = find_packages(),
 	install_requires = [
-		"argh==0.28.1",
+		"argh",
```
ekimekim: This isn't safe, `argh` completely changed their api at some point to be a different library in all but name.
I have an old forked version which I should update to just before the redesign, then we can use that going forward.

HubbeKing: noted! things had a mix of pinned to 0.28.1 and just installing the latest so i assumed it was a python version thing rather than an API thing

ekimekim: ok I just checked their changelog and it's not quite *that* drastic but it's still a lot of breaking changes and they don't seem well communicated.

ekimekim: The last safe version is `0.29.4`

buscribe/setup.py
```diff
@@ -7,10 +7,8 @@
     install_requires = [
         "argh",
         "psycopg2",
-        #"gevent==1.5a2",
-	"gevent",
-        #"greenlet==0.4.16",
-	"greenlet",
+	    "gevent",
+	    "greenlet",
```
ekimekim: now we're no longer pinning `greenlet` we can omit it as it's a `gevent` dependency.
```suggestion
```

api_ping/setup.py
HubbeKing: Buscribe was using latest already, not sure if that's one that also needs to be pinned to 0.28.1

HubbeKing:
Unclear if `downloader` still requires the 7 year old `prometheus-client` version.
I didn't see anything obviously different between `downloader/downloader/main.py` and `playlist_manager/playlist_manager/main.py`, for instance.

ekimekim:
iirc I pinned it so i could reach into internals to do a "set to max(current, new)" operation without needing to manually store a redundant copy of the current value. it probably still works but might break at any time.

HubbeKing:
Alright, all the main images are now "done" - they compile. Things left to look at:
- `nginx` - uses `FROM node:17-alpine AS less` to crunch things before copying into `FROM nginx:latest`
    - node 17 is EOL, probably worth updating to 24, the latest LTS release
- `generate-flamegraph` uses a `FROM alpine:3.7` inline dockerfile - is this even used anymore though?

HubbeKing:
I also had the thought that basically everything uses unpinned python package versions, which means we our image builds are kind of a crapshoot. Ideally we'd pin exact versions and update regularly, but that's a ton of work when we have so many images and they all have separate `setup.py` files.

I dunno if Github's automation for this stuff (DependaBot) can handle having many `setup.py` files in separate directories, but that would be an option. It should open PRs automatically if there's a CVE for anything we have pinned, if configured and enabled. Might be a good thing to try, at least.

ekimekim:
> generate-flamegraph uses a FROM alpine:3.7 inline dockerfile - is this
even used anymore though?

nuke it

PR #600, opened 2026-06-19
Fall back to all-everything playlist thumbnail if no tags match

ekimekim:
This is mainly useful for DBEx where we don't have per-shift templates,
we just want to always use the same template.


chrusher (commented):
Looks fine to me.

PR #601, opened 2026-06-21
downloader: Add max concurrency for segment getters

ekimekim:
Some providers, such as youtube, provide the entire history of segment files.
If we start watching a stream that has already been running for hours, we try to fetch
several thousand segments at once, which causes timeouts at best and gets us blocked at worst.

The solution is a concurrency limit on how many segment getters are running.


chrusher (commented):
Looks good to me.

PR #606, opened 2026-07-11
Thrimbletrimmer SolidJS Rewrite

ElementalAlchemist:
I think this is close enough for a good first submission (let me know if you disagree), so here we go.

Here's the rewrite of Thrimbletrimmer in SolidJS. For a long time, we've talked about how we've reached the complexity limit of what we can reasonably/would like to maintain in raw JavaScript, and that it was time to pick up a framework and have things be more contained and have more inherent connections. This PR aims for feature parity (for which I think it's close enough to submit, but see notes below for differences); further enhancements (which are now either possible or easier) can be added later (more on how we might accomplish that in the future below, too).

## Why SolidJS?
Honestly, mostly personal preference. There are many JS frameworks available (Solid, React, Svelte, Vue, etc.), so ultimately I had to pick one. I didn't like React that much, and Svelte seemed to have a lot of build overhead (though at least some of that may have been my inexperience with getting JS code built). SolidJS felt pretty close to a sweet spot for me; it was reasonable to set up the build, and I like the paradigm since it's similar to what I used with Sycamore and Leptos (frameworks I also liked using).

## What's missing?
There are two things that could be considered regressions:
- I changed the storage format of chapter markers when saving as a draft, mostly to make my life easier. This means that we can no longer round-trip chapter markers for a video that has already been submitted, for which a user is updating the video metadata. We've had plans for a while for various reasons to move chapter marker storage to explicit database fields, and this sort of assumes we'll actually do it, and also hopefully helps push us toward that.
- I'm not sure why the video player doesn't show captions embedded in the video, but so far I haven't been able to get it to do that. I can try to put more time into this as I get time, but if someone knows how to get it to do that, that'd be great. There is support for [text tracks](https://vidstack.io/docs/wc/player/api/text-tracks/?styling=default-theme) if we ever extract or manually create (from any manual transcription efforts) VTT or similar subtitle files.

## What's different?
First, the video player is now based on [Vidstack](https://vidstack.io). Both Vidstack and Thrimbletrimmer still use Hls.js for handling the HLS stream, so our handling of that sort of thing is still the same. However, we get various advantages to using Vidstack:
- Vidstack comes with its own set of icons, so it looks better than all the icons I made in ~2 minutes each in GIMP, and the controls can integrate more easily with the video player without us having to manually hook a bunch of events. (Some of my old icons still live in places that aren't directly related to video playback, but many have been replaced.)
- Vidstack raises more frequent events, which gives us better granularity on timing of various things.
- It also offers the option of limiting playback to a clip, which we might be able to use for clip previews (long requested but we never had granular enough timing to be able to stop the video before; now we could do that or use clips to stop it at the right spot).

Secondly, I replaced JCrop with CropperJS, mostly because I could actually get CropperJS to integrate with SolidJS/my signals. It should work equivalently.

## What can we do?
As a note/reminder, everything in this section is *not in this PR*, but stuff we could do in the future. I at least still want to have this PR with almost-feature-parity because otherwise we could potentially just never submit this.

As I noted above in the section about Vidstack, we could implement the long-requested feature of having a preview for when a clip was ended. (I know this is a bit less important with transitions and the need to have a separate preview for those anyway, but it's there.)

Drag-and-drop reordering of video ranges. Multiple options for this exist; here are two I found a long time ago: [[1]](https://github.com/thisbeyond/solid-dnd) [[2]](https://github.com/isaacHagoel/solid-dnd-directive)

Other enhancements would be available as well; these are just the obvious ones.

## How does it work?
One of the notable technical improvements is that Thrimbletrimmer is now written using TypeScript. Previously, the `nginx` Dockerfile handled copying Thrimbletrimmer files to where nginx could serve them. Now, it adds a step where Thrimbletrimmer is built first. Everything should work fine if you just need it to run.
If you're doing development, use `npm install` to install or update the dependencies, and `npm run build` to run the transpilation. If you're using the dev path option in `docker-compose.jsonnet`, ensure your path ends in the `dist` directory. The example was updated to remind you of this.


thrimbletrimmer/src/editor/Submission.tsx
```diff
@@ -0,0 +1,504 @@
...
+					}
+					if (lastChapterTime !== null && outputVideoTimeSeconds - lastChapterTime < 10) {
+						setSubmissionError("Chapters must be at least 10 seconds apart");
+						props.setEditorState(EditorState.Entry);
+						return;
+					}
+					lastChapterTime = outputVideoTimeSeconds;
+
+					const outputVideoTime = `${Math.floor(outputVideoTimeSeconds / 60)}:${Math.floor(outputVideoTimeSeconds % 60)}`;
+					chapters.push(`${outputVideoTime} - ${chapterData.description}`);
```
dcollinsn, line 146: You mentioned in the PR that chapters don't roundtrip, but do they push to youtube at all? After building `chapters` here, it is unused for the rest of this method.

thrimbletrimmer/src/common/googleAuth.tsx
```diff
@@ -0,0 +1,48 @@
...
+		const signInElem = document.getElementById("google-auth-sign-in");
+		if (signInElem) {
+			signInElem.classList.add("hidden");
+		}
+		const signOutElem = document.getElementById("google-auth-sign-out");
+		if (signOutElem) {
+			signOutElem.classList.remove("hidden");
+		}
+	}
+}
```
dcollinsn, line 33: On sign in, you show the "sign in" button and hide the "sign out" button. Is that backwards?

thrimbletrimmer/src/editor/Submission.tsx
```diff
@@ -0,0 +1,504 @@
...
+			const rangeEnd = rangeEndDate ? wubloaderTimeFromDateTime(rangeEndDate) : null;
+			videoRanges.push([rangeStart, rangeEnd]);
+
+			if (props.chaptersEnabled()) {
+				const chapterDescriptions: string[] = [];
+				for (const chapterData of rangeData.chapters()) {
+					const chapterTime = chapterData.time ? wubloaderTimeFromDateTime(chapterData.time) : null;
+					chapterDescriptions.push(`${rangeDataIndex};${chapterTime} - ${chapterData.description}`);
+				}
+				videoDescription = `${videoDescription}${CHAPTER_MARKER_DELIMITER}${chapterDescriptions.join("\n")}`;
```
dcollinsn: I'm not clear on how this interacts with the code above that processes chapters. However, because this is inside the range data loop, it seems like there will be a "chapter marker delimiter" between chapters belonging to the first range and chapters belonging to the second range, etc?

dcollinsn (commented):
Nits:
* On initial page load, the default is now "Bus Time". The default used to be "Time Ago".
* There are a number of areas of empty space that did not used to be there, for example, above the stream/times bar, and between the player controls and the seek bar, Old vs new:
 
image

image

* The media player seems smaller than it used to be. It now has inline style "--player-width: 1904px; --player-height: 484px; --media-width: 1904px; --media-height: 388px;". It used to be 1:1 at 480p at least.
* "Quality" is now an unlabeled slider, it used to be a dropdown with labeled options "480p" and "source"
* The playback speed used to be visible on the player controls bar, but now is only visible if you click the gear to open the playback control settings.

(I can test live playback using GDQ but I don't think I have the environment set up to test editing or saving as draft)

thrimbletrimmer/src/editor/Submission.tsx
ElementalAlchemist: Whoops, yes, there should only be one delimiter.

thrimbletrimmer/src/editor/Submission.tsx
ElementalAlchemist: ...
welp

thrimbletrimmer/src/common/googleAuth.tsx
ElementalAlchemist: Unfortunately I can't actually test Google auth locally. But yes, this was backward.

thrimbletrimmer/driveclock/drive.html
```diff
@@ -1,24 +0,0 @@
-<!doctype html>
```
ekimekim, line 1: I'd like to keep the driveclock acessible, even if it gets moved elsewhere.
We don't need to keep both versions, just this one.

thrimbletrimmer/vite.config.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export default defineConfig({
+	base: "/thrimbletrimmer/",
+	plugins: [devtools(), solidPlugin()],
+	server: {
+		port: 3000,
+	},
+	build: {
+		target: "esnext",
+		// minify: false, // Uncomment this line if you need to debug unminified code
```
ekimekim: can we please just never minify? it's a massive pain when something goes wrong and the bandwidth savings are negligible.

thrimbletrimmer/thumbnails.html
```diff
@@ -0,0 +1,17 @@
+<!doctype html>
+<html lang="en">
+	<head>
+		<meta charset="utf-8" />
+		<title>Thrimbletrimmer - Utilities</title>
```
ekimekim:
```suggestion
		<title>Thrimbletrimmer - Thumbnail Manager</title>
```

ElementalAlchemist: Whoops, fixed up the original commit

thrimbletrimmer/vite.config.ts
ElementalAlchemist: I thought about this and yeah, fair enough.

thrimbletrimmer/driveclock/drive.html
ElementalAlchemist: OK, so what happened is I did a bunch of stuff near the end of last year and then took a huge break and then recently came back and finished things, and when I came back I assumed I had done the drive clock even though I, in fact, had not.
It's in now.

It turns out I also forgot about the challenges page, though, so I'll do that this weekend.

ElementalAlchemist: I've now added the challenges page. I don't currently have a way to test that one in particular, but it's there. I based it on the existing code so it should work the same.

thrimbletrimmer/src/challenges/Challenges.tsx
```diff
@@ -0,0 +1,108 @@
...
+			const challengesData: ChallengeData[] = await challengesResponse.json();
+			return challengesData;
+		},
+		{ initialValue: [] },
+	);
+
+	return (
+		<>
+			<Show when={challenges() === null}>
+				<div class={styles.loadError}>Failed to get challenges data.</div>
```
ekimekim: any chance we can easily add the error message here?

thrimbletrimmer/src/challenges/Challenges.tsx
```diff
@@ -0,0 +1,108 @@
...
+								body: JSON.stringify({
+									token: authToken,
+									url: url,
+								}),
+							});
+
+							if (submitResponse.ok) {
+								setSubmitResult({ isError: false, message: "Updated successfully." });
+							} else {
+								setSubmitResult({ isError: true, message: "Error during submission" });
```
ekimekim: an error message on how the fetch failed would be helpful here

ekimekim (commented):
I'm slowly working my way through these files. Everything so far is a nitpick or potential future work, not required.

thrimbletrimmer/src/common/convertTime.tsx
```diff
@@ -0,0 +1,129 @@
...
+	}
+	return diff.toFormat("hh:mm:ss.SSS");
+}
+
+export function dateTimeFromTimeAgo(timeAgo: string): DateTime | null {
+	const parts = timeAgo.split(":");
+	const properties = ["hours", "minutes", "seconds"];
+	const mathObj = {};
+
+	while (parts.length > 0) {
```
ekimekim, line 72: Isn't this the same parsing process (`HH:MM:SS` -> math object) as for bustime? The implementations are shaped quite differently which could be inconsistent.

thrimbletrimmer/src/common/convertTime.tsx
```diff
@@ -0,0 +1,129 @@
...
+	fragments: Fragment[],
+	videoTime: number,
+): DateTime | null {
+	let fragmentStartTime: number | undefined = undefined;
+	let fragmentStartISOTime: string | undefined = undefined;
+	for (const fragment of fragments) {
+		const fragmentEndTime = fragment.start + fragment.duration;
+		if (videoTime >= fragment.start && videoTime < fragmentEndTime) {
+			fragmentStartTime = fragment.start;
+			fragmentStartISOTime = fragment.rawProgramDateTime!;
```
ekimekim, line 119: The `!` operator always makes me uneasy (it's morally equivalent to rust's `unsafe`, except way harder to spot).
In previous typescript codebases I've worked with we had the following helper function instead:
```
function defined<T>(value: T | null | undefined): T {
  if (value === null || value === undefined) {
    throw new Error(`Value asserted as defined is ${value}`);
  }
  return value;
}
```
(usage example: `fragmentStartISOTime = defined(fragment.rawProgramDateTime)`)
so that you get a crash instead of silently mistyped values.

If nothing else, I'd prefer `value as string` over `value!` just because it's more visible and obvious you're doing a type assertion.

thrimbletrimmer/src/common/downloads.tsx
```diff
@@ -0,0 +1,6 @@
+export enum DownloadType {
+	Smart,
+	Rough,
+	Fast,
```
ekimekim: tbh I think we can drop the "fast" option. The "smart" option completely subsumes it and has been in use for long enough now that we fully trust it.

dcollinsn:
I am not sure if the comments I made in https://github.com/dbvideostriketeam/wubloader/pull/606#pullrequestreview-4676703846 were handled. However, I don't have any further comments beyond that, and would endorse merging this so that work dependent on the decision of whether to merge this (such as #610) can move forwards.

ElementalAlchemist:
> I am not sure if the comments I made in [#606 (review)](https://github.com/dbvideostriketeam/wubloader/pull/606#pullrequestreview-4676703846) were handled. However, I don't have any further comments beyond that, and would endorse merging this so that work dependent on the decision of whether to merge this (such as #610) can move forwards.

I think I missed that comment earlier, but I can try to address things:



> Nits:
> 
> * On initial page load, the default is now "Bus Time". The default used to be "Time Ago".

Yes, but it still defaults to the same range (starting 10 minutes ago). Is this a problem, or should we use this opportunity to change the default? IMO "time ago" is mostly not useful beyond the default load (unless you're just adjusting that time, but we have the pad buttons).

> * There are a number of areas of empty space that did not used to be there, for example, above the stream/times bar, and between the player controls and the seek bar, Old vs new:
> 
> image image

I could potentially try to futz with some of the controls a bit, but I didn't make the new controls (on purpose) so we'd need to see. I was sort of on the side of "we shouldn't customize vidstack's provided controls too much".

> * The media player seems smaller than it used to be. It now has inline style "--player-width: 1904px; --player-height: 484px; --media-width: 1904px; --media-height: 388px;". It used to be 1:1 at 480p at least.

This is probably also on vidstack's end. I think I'm only applying the same styling as before, which is that the maximum height of the player is half of the window height.

> * "Quality" is now an unlabeled slider, it used to be a dropdown with labeled options "480p" and "source"

This is probably a little bit of both; if you can find a working quality selector control for Vidstack I can put in the same spot that you'd prefer, feel free to submit that and we'll use it.

> * The playback speed used to be visible on the player controls bar, but now is only visible if you click the gear to open the playback control settings.

Previously nothing was in a menu. If this gets used more frequently than I think, we can look into moving it, but I don't remember if having a slider outside the menu works as well as just putting it where it is.

> 
> (I can test live playback using GDQ but I don't think I have the environment set up to test editing or saving as draft)

ekimekim:
> If this gets used more frequently than I think

When I'm editing I'm shifting playback speeds constantly (mostly using the keyboard shortcuts) and being able to see the current speed is important. I know a lot of other editors do the same, editing at 1.5x or faster then slowing down to find a good cut point.

ElementalAlchemist:
If you're not using the slider itself, I can just plop the playback rate somewhere, and that could eliminate worries about the placement of the control.

thrimbletrimmer/src/common/convertTime.tsx
ElementalAlchemist: Sort of? They're similar, but (for whatever reason) bus time builds from hours (so `6` is 6 hours, `6:30` is 6 hours and 30 minutes, and `6:30:21` is 6 hours, 30 minutes, and 21 seconds) whereas time ago builds from seconds (so `6` is 6 seconds, `6:30` is 6 minutes and 30 seconds, and `6:30:21` is 6 hours, 30 minutes, and 21 seconds).

thrimbletrimmer/src/common/convertTime.tsx
ElementalAlchemist: That's fair but
1. I'm used to looking for exclamation points (which, yes, I know is a me thing, and not conducive reasoning for a codebase where we need to work together)
2. In this specific case, we know `rawProgramDateTime` is defined (otherwise many other parts of Thrimbletrimmer also break, and the same is true pre-rewrite), and
3. If we're going to do that, we should find all the type-coercion `!` uses, which are more than just this one.

If you'd like me to try to go through the exercise, I can make an attempt at doing that exercise at some point (but it'll be a bit).

thrimbletrimmer/src/challenges/Challenges.tsx
ElementalAlchemist: Added.

thrimbletrimmer/src/challenges/Challenges.tsx
ElementalAlchemist: Added.

thrimbletrimmer/src/common/downloads.tsx
ElementalAlchemist: Removed from the UI. I'll leave you in charge of killing it in the backend.

ElementalAlchemist:
I started trying to move the control when I realized I could just show it in the controls bar next to the menu gear (so it's part of the menu open button).

ekimekim (approved):
Realistically, I am not going to get through all this code with a fine-toothed comb. I am happy to merge as-is and if there are issues we can address them as we hit them.

ElementalAlchemist:
Given
- ekim is good to merge for now
- Sense reviewed this a bit and then seems to have disappeared
- we need this for our main test of it starting tomorrow
I'm going to go ahead and merge this now. If you have further comments on this, please write issues or hit me up on Zulip.

PR #608, opened 2026-07-11
restreamer: trim holes within last 15 seconds of live

### LINKED ISSUE #607: Restreamer in live mode makes non-standards-compliant changes to playlist when segments are downloaded out of order
When playing in restreamer without an end time ("live mode"), if a later segment is present and an earlier segment is missing, the restreamer appends an `#EXT-X-DISCONTINUITY` tag in place of the missing segment, leaving for example the following playlist:

```m3u8
#EXT-X-PROGRAM-DATE-TIME:2026-07-11T16:56:41.001000Z
#EXTINF:2.000,live
...56%3A41.001000...ts
#EXT-X-DISCONTINUITY
#EXT-X-PROGRAM-DATE-TIME:2026-07-11T16:56:45.001000Z
#EXTINF:2.000,live
...56%3A45.001000...ts
```

If that missing segment is later fetched, future loads of that playlist will have that segment:

```m3u8
#EXT-X-PROGRAM-DATE-TIME:2026-07-11T16:56:41.001000Z
#EXTINF:2.000,live
...56%3A41.001000...ts
#EXT-X-PROGRAM-DATE-TIME:2026-07-11T16:56:43.001000Z
#EXTINF:2.000,live
...56%3A43.001000...ts
#EXT-X-PROGRAM-DATE-TIME:2026-07-11T16:56:45.001000Z
#EXTINF:2.000,live
...56%3A45.001000...ts
```

This does not comply with [RFC8216](https://datatracker.ietf.org/doc/html/rfc8216#section-6.2.1):

> The server MUST NOT change the Media Playlist file, except to:
> 
>    o  Append lines to it ([Section 6.2.1](https://datatracker.ietf.org/doc/html/rfc8216#section-6.2.1)).
> 
>    o  Remove Media Segment URIs from the Playlist in the order that they
>       appear, along with any tags that apply only to those segments
>       ([Section 6.2.2](https://datatracker.ietf.org/doc/html/rfc8216#section-6.2.2)).
> 
>    o  Increment the value of the EXT-X-MEDIA-SEQUENCE or EXT-X-
>       DISCONTINUITY-SEQUENCE tags ([Section 6.2.2](https://datatracker.ietf.org/doc/html/rfc8216#section-6.2.2)).
> 
>    o  Add an EXT-X-ENDLIST tag to the Playlist ([Section 6.2.1](https://datatracker.ietf.org/doc/html/rfc8216#section-6.2.1)).

Under the current JS player, I would occasionally see repeated or skipped segments, and cases where the audio of one segment would play over the video of a different segment, but I wasn't able to figure out why. While testing the SolidJS version, I would occasionally have playback completely stall. I noticed in dev tools that this happened immediately after the following sequence:

1. Fetch of an m3u8 playlist where the second to last media segment is missing
2. Fetch of the last media segment
3. Fetch of an m3u8 playlist which fills that discontinuity with the second to last segment
4. Fetch of the new media segment

At this point, the player would permanently stall, with no console warnings, it would stop trying to fetch the playlist. 

My proposal to fix this is, during live playback (playlist fetch with no end time), if there is a discontinuity within the last N seconds, then end the playlist before the discontinuity. in the hopes that the segment will be fetched for inclusion in the next playlist.

dcollinsn:
This patch updates the restreamer to trim up to the last 15 seconds from a playlist if a hole appears within that time period.

This normally occurs when the downloader fetches segments out of order. If the missing segment is later fetched, from the live client's point of view, it will be added to the playlist and not at the end. This violates RFC8216 and causes player anomalies.

Trimming the holes and subsequent segments gives the downloader up to 15 seconds to fetch the missing segments, preventing this issue.

Resolves #607

### DIFF
--- a/restreamer/restreamer/main.py
+++ b/restreamer/restreamer/main.py
@@ -188,6 +188,47 @@ def time_range_for_quality(channel, quality):
 	return parse_hour(first), parse_hour(last) + datetime.timedelta(hours=1)
 
 
+LIVE_TAIL_HOLE_GRACE = datetime.timedelta(seconds=15)
+
+
+def trim_live_tail_holes(segments):
+	"""Do not publish segments after small and recent holes in open-ended live playlists.
+
+	Segments can arrive slightly out of order near the live edge. If we publish a
+	later segment across that hole, and then later receive the missing segment,
+	the playlist will change in a non-compliant way and the player can stall or
+	glitch.
+	"""
+	segments = list(segments)
+	# Find the index of the last real segment, which we need for some checks later
+	last_segment_index = next(
+		(i for i in reversed(range(len(segments))) if segments[i] is not None),
+		None,
+	)
+	if last_segment_index is None:
+		# There are no real segments in this array, so just pass it through
+		return segments
+
+	# Ending timestamp of the latest available segments. This is the timestamp
+	# from which we calculate the grace period for out of order downloading.
+	live_tail = segments[last_segment_index].end
+
+	for i, segment in enumerate(segments):
+		# Look for the start of a "hole" of missing segments
+		if segment is not None:
+			continue
+		if i == 0 or segments[i - 1] is None:
+			continue
+
+		# Calculate the time from the start of this hole to the end of the last
+		# segment. If this hole is within the grace period of live, then end the
+		# playlist immediately before this hole.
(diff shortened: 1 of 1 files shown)


ekimekim (approved):
Annoying, but I don't see a better fix.

PR #623, opened 2026-08-24
backend: Manage chapter markers as a database column

### LINKED ISSUE #368: Thrimbletrimmer: Store in-progress chapter markers better
From a discussion in Zulip today: It'd be better to be able to store chapter markers more discretely to prevent potential conversion issues between the description-stored values and the actual values, only generating the description once when shipping to YouTube.

We haven't encountered issues we can definitively point to this, but it may make things generally neater.

ekimekim:
instead of the editor automatically appending them to the description.

This adds an edit column "chapter_markers" containing a list of (time, name).
The time is a time interval from video start (eg. 1 minute 23 seconds) and the name is the chapter name.

The chapter markers are appended to the description automatically during cutting,
and can be modified by the usual method.

### DIFF
--- a/cutter/cutter/main.py
+++ b/cutter/cutter/main.py
@@ -8,6 +8,7 @@ import random
 import signal
 import socket
 from collections import namedtuple
+from itertools import zip_longest
 
 import gevent.backdoor
 import gevent.event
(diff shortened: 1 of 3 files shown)


ElementalAlchemist:
Part of the point of changing the chapter marker format away from this was to avoid some of the round-tripping problems we experienced in the old system, which would be required by using the published video time format. (The old system included an extra jump between the published video time and a human-readable version, but that was an easy transformation and not the buggy part.)

The looser validation when saving drafts (at least in Thrimbletrimmer, if not thrimshim) means that, nominally, some of the ranges could be open-ended at the time of saving a draft, which would make determining a published video time impossible. (The same is not true when submitting, as all ranges must have defined start and end times by that point.)

ekimekim:
Ok then, the tradeoff will be that manually uploaded videos cannot be updated with chapter markers since we don't know their video ranges...but I guess that was true on the frontend already.

ElementalAlchemist (approved):

PR #624, opened 2026-08-25
thrimshim: Relax validation for unedited videos

### LINKED ISSUE #378: thrimbletrimmer: Relax validation when saving as draft
When saving as draft, it is not nessecary that all the checks we have for a well-formed video need to be valid.
For example, it should be possible to not yet have an end time. A common use-case for this is editing the start of a video before it has finished.

We should evaluate each validation check and decide if it should only be enforced on actual submissions.

Some of these checks might also need to be relaxed on the thrimshim end.

ekimekim:
This allows edit columns to be null and remain unvalidated, as long as the state is being set to UNEDITED.
The following exceptions apply:
- Title and description must be non-null (though may be empty), and max length limits remain enforced
- If video_ranges is set, it must be non-empty and video_transitions must be valid.

### DIFF
--- a/thrimshim/thrimshim/main.py
+++ b/thrimshim/thrimshim/main.py
@@ -335,13 +335,16 @@ def update_row(ident, editor=None):
 	new_row = flask.request.json
 	override_changes = new_row.get('override_changes', False)
 	state_columns = ['state', 'uploader', 'error', 'video_link']
-	# These have to be set before a video can be set as 'EDITED'
+	# These must always be set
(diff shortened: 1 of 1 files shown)


ElementalAlchemist:
Notes on those exceptions:
- I currently pass title and description as `string` (not `string | null`) so we should be alright on that
- I'll need to do an `if videoRanges.length === 0 { videoRanges = null }` to make that works (but that's doable, so it's fine)

PR #627, opened 2026-09-08
Dobby/sup reminders

### LINKED ISSUE #503: Reminder bot for shift sups
Sups have requested certain periodic reminders for themselves in Zulip, such as "tell people to take breaks".
This should be a very simple bot on a timer, maybe built into schedulebot?

Dobbykroket:
Initial implementation of reminders in schedulebot, resolves #503 

Reminders are added in the docker-compose.jsonnet file, in the args array:
`"--reminder", "general:reminders:1,3,5,7,9,11,13,15,17,19,21,23:Hydrate everybody!", "general:reminders:0:Jock Jams!"`

A reminder is formated as `STREAM:TOPIC:TIMESPEC:TEXT`, with timespec being a comma separated list of moonbase hours [0-23].

Possible further improvements:
Allowing reminders to be triggered at times outside a full hour (maybe switch timespec to a cron-like format?)
maybe cleaning up the reminder accessing in check_reminders so that we're not accessing "random" indexes, but I'm not really sure about a neat way to do that in python.

### DIFF
--- a/zulip_bots/zulip_bots/schedulebot.py
+++ b/zulip_bots/zulip_bots/schedulebot.py
@@ -76,7 +76,7 @@ def update_groups(client, group_ids, schedule, hour, start_time, last):
 
 def hour_to_shift(hour, start_time, shift_definitions):
 	"""Converts an hour number into a datetime, shift emoji name, and hour-of-shift (1-6)"""
-	start_time = datetime.utcfromtimestamp(start_time)
+	start_time = datetime.fromtimestamp(start_time, timezone("utc"))
 	current_time = (start_time + timedelta(hours=hour)).replace(minute=0, second=0, microsecond=0)
 	shifts = parse_shifts(shift_definitions) # need to do this each hour to check for Omega shift start time
 	shift, shift_hour = calculate_shift(current_time, shifts)
--- a/zulip_bots/zulip_bots/schedulebot.py
+++ b/zulip_bots/zulip_bots/schedulebot.py
@@ -17,6 +17,7 @@ import time
 from calendar import timegm
 from datetime import datetime, timedelta
 from pytz import timezone
+from collections import namedtuple
 
 import gevent.pool
 import argh
@@ -245,22 +246,27 @@ def parse_schedule(sheets_client, user_ids, schedule_sheet_id, schedule_sheet_na
 	return schedule
 
 
+Reminder = namedtuple("Reminder", ["times", "stream", "topic", "text"])
 def parse_reminders(reminder_strings: list[str]):
 	reminders = list()
 	for reminder_string in reminder_strings:
 		stream, topic, timespec, text = reminder_string.split(':', 3)
 		times = [int(t) for t in timespec.split(',')]
-		reminders.append((times, stream, topic, text))
+		reminders.append(Reminder(times, stream, topic, text))
 	return reminders
 
 
-def check_reminders(reminders, send_client):
-	moonbase_hour = datetime.now(timezone("America/Vancouver")).hour
+def check_reminders(reminders: list[Reminder], hour, start_time, send_client):
+	if hour < 0: return # Don't send reminders before the run starts
+	moonbase_start = datetime.fromtimestamp(start_time, timezone("America/Vancouver"))
+	moonbase_hour = (moonbase_start + timedelta(hours=hour)).replace(minute=0, second=0, microsecond=0).hour
 	for reminder in reminders:
-		if moonbase_hour in reminder[0]:
-			send_client.send_to_stream(reminder[1], reminder[2], reminder[3])
+		if moonbase_hour in reminder.times:
+			send_client.send_to_stream(reminder.stream, reminder.topic, reminder.text)
 
 
+
+@argh.arg("--reminder", action="append", metavar="STREAM:TOPIC:HOUR{,HOUR}:TEXT")
 def main(conf_file, hour=-1, no_groups=False, stream="General", no_mentions=False, no_initial=False, shifts=None, last=-1, metrics_port=8012, reminder: list[str] = None):
 	"""
 	config:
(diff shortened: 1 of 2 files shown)


zulip_bots/zulip_bots/schedulebot.py
```diff
@@ -244,7 +245,23 @@ def parse_schedule(sheets_client, user_ids, schedule_sheet_id, schedule_sheet_na
...
+
+
+def check_reminders(reminders, send_client):
+	moonbase_hour = datetime.now(timezone("America/Vancouver")).hour
+	for reminder in reminders:
+		if moonbase_hour in reminder[0]:
+			send_client.send_to_stream(reminder[1], reminder[2], reminder[3])
+
+
+def main(conf_file, hour=-1, no_groups=False, stream="General", no_mentions=False, no_initial=False, shifts=None, last=-1, metrics_port=8012, reminder: list[str] = None):
```
ekimekim, line 270: We'll need an arg specifier on this to make sure argh knows to accept it multiple times. We can also document the expected format while we're there.

```suggestion
@argh.arg("--reminder", action="append", metavar="STREAM:TOPIC:HOUR{,HOUR}:TEXT")
def main(conf_file, hour=-1, no_groups=False, stream="General", no_mentions=False, no_initial=False, shifts=None, last=-1, metrics_port=8012, reminder = None):
```

zulip_bots/zulip_bots/schedulebot.py
```diff
@@ -289,6 +306,7 @@ def main(conf_file, hour=-1, no_groups=False, stream="General", no_mentions=Fals
 		config["schedule_sheet_id"],
 		config["schedule_sheet_name"]
 	)
+	reminders = parse_reminders(reminder)
```
ekimekim: `reminder` will be `None` if no `--reminder` args are given, not `[]`, so we need to account for that:
```suggestion
	reminders = parse_reminders(reminder or [])
```

zulip_bots/zulip_bots/schedulebot.py
```diff
@@ -244,7 +245,23 @@ def parse_schedule(sheets_client, user_ids, schedule_sheet_id, schedule_sheet_na
...
+	reminders = list()
+	for reminder_string in reminder_strings:
+		stream, topic, timespec, text = reminder_string.split(':', 3)
+		times = [int(t) for t in timespec.split(',')]
+		reminders.append((times, stream, topic, text))
+	return reminders
+
+
+def check_reminders(reminders, send_client):
+	moonbase_hour = datetime.now(timezone("America/Vancouver")).hour
```
ekimekim: `check_reminders` should take a passed-in current `hour` rather than determining it on its own (it will also need to know `start_time`) - this is important for testing where we explicitly pass a `--hour` to use.

zulip_bots/zulip_bots/schedulebot.py
```diff
@@ -244,7 +245,23 @@ def parse_schedule(sheets_client, user_ids, schedule_sheet_id, schedule_sheet_na
 	return schedule
 
 
-def main(conf_file, hour=-1, no_groups=False, stream="General", no_mentions=False, no_initial=False, shifts=None, last=-1, metrics_port=8012):
+def parse_reminders(reminder_strings: list[str]):
+	reminders = list()
+	for reminder_string in reminder_strings:
+		stream, topic, timespec, text = reminder_string.split(':', 3)
+		times = [int(t) for t in timespec.split(',')]
+		reminders.append((times, stream, topic, text))
```
ekimekim: > maybe cleaning up the reminder accessing in check_reminders so that we're not accessing "random" indexes, but I'm not really sure about a neat way to do that in python.

The neat python thing for this is `collections.NamedTuple`:
```
Reminder = namedtuple("Reminder", ["times", "stream", "topic", "text"])
```
```
reminders.append(Reminder(times, stream, topic, text))
```
```
send_client.send_to_stream(reminder.stream, reminder.topic, reminder.text)
```

ekimekim (commented):
A few bits to fix, but the approach looks good.
Could you also add support to `docker_compose.jsonnet` to specify the reminders in the config? So that declaring reminders looks like this:
```jsonnet
reminders:: [
  {
    stream: "#general",
    topic: "Important reminders",
    hours: [0, 12],
    text: ```
      It's time to...
      ```spoiler Time to what?
      **ROTATE THE CUSHIONS**
      ```
    ```,
  },
],
```
Sorry to dump you into a wholly unfamiliar language but there should be plenty of examples to work from. The main line you need should be something like:
```jsonnet
  // other schedulebot args...
] + [
  "--reminder=%s:%s:%s:%s" % [
    r.stream,
    r.topic,
    std.join(",", std.map(std.toString, r.hours)),
    r.text,
  ]
  for r in $.reminders
]
```

Dobbykroket:
putting it in to the config instead having it as a command line arg makes a lot of sense, I'll take a look at how that works.

zulip_bots/zulip_bots/schedulebot.py
Dobbykroket: Ah, that's how you're supposed to do it, I couldn't find it in the docs.

It looks like they'll be depreciating the decorator at some point, but that's a future issue.

ekimekim:
It's actually both! The `docker-compose.jsonnet` is the central config file across services, but it works by specifying CLI args for each service. So the CLI arg is the transfer medium (as well as being useful for testing).

zulip_bots/zulip_bots/schedulebot.py
ekimekim: Yeah we're already pinning to an older version for other reasons, they want to move the library in a completely different direction and break everything.

Dobbykroket:
Yeah I saw, I mostly meant specifying it in the config rather than bolting it on to the config as a CLI arg

zulip_bots/zulip_bots/schedulebot.py
Dobbykroket: ah, the classic problem... _fun_

Dobbykroket:
Implemented all suggestions.

Also updated a depreciated function call while I was playing around with datetimes anyway

ekimekim (approved):

PR #628, opened 2026-09-14
Port bots to new website APIs

ekimekim:
Adds nice interfaces for the new website, for both REST and streaming APIs.
Converts the various bots to use it, including the new streaming apis where relevant.
See individual commits for more detail.


chrusher (approved):
Looks ok to me

